PR #240, opened 2023-06-17
feat: remove CreateTable and friends

ajewellamz:
*Issue #, if available:*

*Description of changes:*


texastony (commented):
Question: I feel like I am missing some context.
Have we determined that we do not need to modify Table requests?

ajewellamz:
That is correct. Most customers will want to use the console or CloudFormation to create their tables and indexes, and so patching the API would have a fairly small audience, and we would still need to document how do do it manually.

Given that the workshop was ok without it, it was determined that the best approach was to remove it altogether, rather than to add the java code to enable it.

PR #246, opened 2023-07-10
feat: show multi-threading in example

ajewellamz:
*Issue #, if available:*

*Description of changes:*


Examples/runtimes/java/DynamoDbEncryption/src/main/java/software/amazon/cryptography/examples/searchableencryption/CompoundBeaconSearchableEncryptionExample.java
```diff
@@ -232,6 +232,23 @@ public static void PutItemQueryItemWithCompoundBeacon(String ddbTableName, Strin
...
+    // it might look like this
+    Runnable myThread = () -> {
+        for(int i = 0; i < 20; ++i)
+            QueryItemWithCompoundBeacon(ddb, ddbTableName);
+    };
+    // increase once we expect threads to work
+    for(int i = 0; i < 1; ++i) {
+        Thread run = new Thread(myThread);
+    }
+  }
```
lucasmcdonald3, line 248: I think a thread pool would be cleaner here:

```
import java.util.concurrent.ExecutorService;
import java.util.concurrent.Executors;

...

MAX_CONCURRENT_QUERY_THREADS = 1;
```

Then,

```suggestion
    // If instead you were working in a multi-threaded context
    // it might look like this
    Runnable myThread = () -> {
        for (int i = 0; i < 20; i++) {
            QueryItemWithCompoundBeacon(ddb, ddbTableName);
        }
    };
    ExecutorService pool = Executors.newFixedThreadPool(MAX_CONCURRENT_QUERY_THREADS);  
    for(int i = 0; i < MAX_CONCURRENT_QUERY_THREADS; i++) {
        pool.execute(myThread);
    }
    pool.shutDown();
```

We could also manage threads individually, but I think we would have to manage them in some data structure to ensure all threads `join` before we return success from the test, e.g.

```
    Thread[] threads = new Thread[MAX_CONCURRENT_QUERY_THREADS];
    // If instead you were working in a multi-threaded context
    // it might look like this
    Runnable myThread = () -> {
        for(int i = 0; i < 20; i++)
            QueryItemWithCompoundBeacon(ddb, ddbTableName);
    };
    // increase once we expect threads to work
    for(int i = 0; i < MAX_CONCURRENT_QUERY_THREADS; i++) {
        Thread run = new Thread(myThread);
        threads[i] = run;
        run.start();
    }
    for(int i = 0; i < MAX_CONCURRENT_QUERY_THREADS; i++) {
        run.join()
    }
```

PR #255, opened 2023-07-16
test(enhancedClient): Verify EC's null behavior does not interfere

texastony:
*Issue #, if available:*

*Description of changes:*
- Add Lombok as a test dependency
- Use Lombok to create a DDB Data Model with a `lastModifiedTimeStamp`. 
- Use this Data Model to verify that the Enhanced Client's (EC) Null Behavior 
does not interfere with the DB-ESDK for DynamoDB

Note: Authoring this test is how I discovered the issue with Index Member identification,
which is fixed in https://github.com/aws/aws-database-encryption-sdk-dynamodb-java/pull/254


ajewellamz:
No longer necessary.

PR #296, opened 2023-07-29
Changed type to N and added consistent read to true

LeeroyHannigan:
- The comments wrongly state that sort key is of type (S), whereas it is type (N), updated the comment.
- Added strong consistency for the read, to ensure you retrieve the item directly after writing.

### DIFF
--- a/Examples/runtimes/java/DynamoDbEncryption/src/main/java/software/amazon/cryptography/examples/BasicPutGetExample.java
+++ b/Examples/runtimes/java/DynamoDbEncryption/src/main/java/software/amazon/cryptography/examples/BasicPutGetExample.java
@@ -27,7 +27,7 @@ import java.util.Map;
   This table must be configured with the following
   primary key configuration:
     - Partition key is named "partition_key" with type (S)
-    - Sort key is named "sort_key" with type (S)
+    - Sort key is named "sort_key" with type (N)
  */
 public class BasicPutGetExample {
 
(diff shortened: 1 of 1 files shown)


Examples/runtimes/java/DynamoDbEncryption/src/main/java/software/amazon/cryptography/examples/BasicPutGetExample.java
```diff
@@ -154,6 +154,7 @@ public static void PutItemGetItem(String kmsKeyId, String ddbTableName) {
         final GetItemRequest getRequest = GetItemRequest.builder()
                 .key(keyToGet)
                 .tableName(ddbTableName)
+                .consistentRead(Boolean.TRUE)
```
lavaleri, line 157:
```suggestion
                // In this example we configure a strongly consistent read
                // because we perform a read immediately after a write (for demonstrative purposes).
                // By default, reads are only eventually consistent.
                // Read our docs to determine which read consistency to use for your application:
                // https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.ReadConsistency.html
                .consistentRead(Boolean.TRUE)
```

lavaleri (changes requested):
Thanks for the PR @LeeroyHannigan!

Adding a consistent read to this example is a great update. In order to make it clear that we aren't trying to be prescriptive around whether one should or should not use consistent reads in general, I have suggested adding a comment alongside your change in the PR.

LeeroyHannigan:
@lavaleri  thanks for that. I noticed other parts of the examples will require the same changes, not sure if you know of an easy way to implement it in all needed files?

lavaleri:
Thanks for the update! Our CI isn't set up currently to accept changes from forks, so I have pushed this change to a branch directly off of upstream (and closed this PR): https://github.com/aws/aws-database-encryption-sdk-dynamodb-java/pull/298

It appears that there may be another permission issue affecting our CI right now that we will look into. We'll keep the changes in that PR scoped to just the simple example for now.

PR #422, opened 2023-09-15
chore: Always test/build with locally built MPL and add release safeg…

lavaleri:
…uards

Updates CI to build/test using local MPL, making it easier to reason about local development, and unblock development with pre-release MPL features.

We already ensure in our release process that we test the examples and test vectors against MPL in maven local, but this PR adds an additional script to ensure there is consistency between the MPL submodule and the MPL version in our build.gradle that can be used before and during a release. It isn't checked in CI, in order to support development against pre-release MPL.

Updated the build.gradle to use the same versions of MPL/Dafnyruntime/etc. across the DB-ESDK, examples, and test vectors. These versions are currently defined in a top level `properties` file, that each individual gradle project reads from. This is Java specific for now, to give us the best bang for our buck today.

Once effect of this is that I think these dependencies will no longer be caught by dependabot. This is probably ok, as we want to be mindful when upgrading these dependencies anyway.

Tested the Codebuild script update on commits [94fc4b4](https://github.com/aws/aws-database-encryption-sdk-dynamodb-java/pull/422/commits/94fc4b46b6b66ab1c1e0c544ce40d4722d92e664), [79926a4](https://github.com/aws/aws-database-encryption-sdk-dynamodb-java/pull/422/commits/79926a404ed9b10ceb76ef19c05f130f59022019), and [e219262](https://github.com/aws/aws-database-encryption-sdk-dynamodb-java/pull/422/commits/e219262b439b6f0692c5bd56db9ab5396d1437c4) to ensure expected success/failures.

### DIFF
--- a/codebuild/staging/release-staging.yml
+++ b/codebuild/staging/release-staging.yml
@@ -47,8 +47,10 @@ phases:
       - aws sts get-caller-identity
   build:
     commands:
-      - cd DynamoDbEncryption/
+      # Validate the MPL submodule points to the correct release
+      - scripts/validate-mpl-submodule.sh
       # Build and deploy to maven local
+      - cd DynamoDbEncryption/
       - make transpile_implementation_java
       - make transpile_test_java
       - make mvn_local_deploy
--- a/project.properties
+++ b/project.properties
@@ -1,0 +1,4 @@
+projectJavaVersion=3.1.0
+mplDependencyJavaVersion=1.0.0
+dafnyRuntimeJavaVersion=4.1.0
+smithyDafnyJavaConversionVersion=0.1
--- a/project.properties
+++ b/project.properties
@@ -1,4 +1,4 @@
-projectJavaVersion=3.1.0-SNAPSHOT
+projectJavaVersion=3.1.0
 mplDependencyJavaVersion=1.0.0
 dafnyRuntimeJavaVersion=4.1.0
 smithyDafnyJavaConversionVersion=0.1
--- a/project.properties
+++ b/project.properties
@@ -1,4 +1,4 @@
-projectJavaVersion=3.1.0
+projectJavaVersion=3.1.0-SNAPSHOT
 mplDependencyJavaVersion=1.0.0
 dafnyRuntimeJavaVersion=4.1.0
 smithyDafnyJavaConversionVersion=0.1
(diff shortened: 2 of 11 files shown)


project.properties
```diff
@@ -0,0 +1,4 @@
+projectJavaVersion=3.1.0
```
josecorella, line 1: nit: should this be dbesdkJavaVersion instead?

codebuild/staging/release-staging.yml
```diff
@@ -47,8 +47,10 @@ phases:
       - aws sts get-caller-identity
   build:
     commands:
-      - cd DynamoDbEncryption/
+      # Validate the MPL submodule points to the correct release
+      - scripts/validate-mpl-submodule.sh
```
texastony, line 51: Question/Issue: 
The Script validates that the MPL Submodule is at the same tag 
as the declared dependency.

But this build, test, and deploy does NOT use the MPL from Maven.
It uses the MPL from the Submodule, correct?

During the Release process, 
is the library ever tested with the MPL from Maven?

If so, is this test before or after the push to Maven?

ajewellamz (approved):
LGTM

codebuild/staging/release-staging.yml
lavaleri: > During the Release process,
is the library ever tested with the MPL from Maven?

Yes, the MPL isn't getting built in the release scripts right now. Note that we are using `transpile` which only transpiles the current project, instead of `build` which will recursively build everything.

staging test: https://github.com/aws/aws-database-encryption-sdk-dynamodb-java/blob/main/codebuild/staging/release-staging.yml#L51-L55
staging examples/test vectors: https://github.com/aws/aws-database-encryption-sdk-dynamodb-java/blob/main/codebuild/staging/validate-staging.yml#L43-L65
release test: https://github.com/aws/aws-database-encryption-sdk-dynamodb-java/blob/main/codebuild/release/release-prod.yml#L53-L57

The fact that our `build` target automatically recursively builds everything is a sharp edge that we should eventually fix.

josecorella (approved):
left a small nit but LGTM

texastony (approved):

project.properties
lavaleri: ah, I was thinking that it would be more obvious that this meant "this project". I.e. if the MPL implemented a similar thing, it would use `projectJavaVersion` to mean what Java version of the _MPL_ we're on.

PR #459, opened 2023-09-28
feat: dont use internal, because it is reserved in dotnet

ajewellamz:
*Issue #, if available:*

*Description of changes:*


texastony (changes requested):
This is not idiomatic Java.
Let's talk to Jose and Val about this.
We may need to restructure this repo like Jose did for .NET.

texastony (approved):

seebees:
Late but LGTM

PR #466, opened 2023-10-01
feat: repair dotnet build

ajewellamz:
*Issue #, if available:*

*Description of changes:*


DynamoDbEncryption/runtimes/net/Generated/DynamoDbEncryption/AsSet.cs
texastony, line 1: Question: since this adds .NET to this repo,
we could format our .NET NOW,
and make all of this readable.

Would you be willing to add that to the scope of this PR?
I have code for this.

DynamoDbEncryption/runtimes/net/Generated/DynamoDbEncryption/AsSet.cs
```diff
@@ -0,0 +1,13 @@
...
+using System;
+ using AWS.Cryptography.DbEncryptionSDK.DynamoDb; namespace AWS.Cryptography.DbEncryptionSDK.DynamoDb {
+ public class AsSet {
+
+ 
+ public void Validate() {
+ 
+}
+}
+}
```
texastony, line 13: Question: Does this really work?
I was comparing [it to the Java code](https://github.com/aws/aws-database-encryption-sdk-dynamodb-java/blob/main/DynamoDbEncryption/runtimes/java/src/main/smithy-generated/software/amazon/cryptography/dbencryptionsdk/dynamodb/model/AsSet.java),
which I am now also suspicious of.

texastony (commented):
There are a lot of files to review.
I left two questions.
Will come back and review more later.

DynamoDbEncryption/runtimes/net/Generated/DynamoDbEncryption/AsSet.cs
ajewellamz: I ran `make format_net` even though I'm not sure this is getting merged.
Also ran the new polymorph, so we only have one manual edit left.

DynamoDbEncryption/runtimes/net/Generated/DynamoDbEncryption/AsSet.cs
ajewellamz: It should. It has no members and no behavior --it's just a way of selecting one option from a union.

Of what are you suspicious?

texastony: It... it looks' logic-less, and therefore pointless?
But maybe I am missing something.

ajewellamz: Look in the .smithy file, I think that will clear it up. It has
`structure AsSet {}`
so that we can have
```
union BeaconStyle {
  partOnly: PartOnly,
  twinned: Twinned,
  asSet: AsSet,
  twinnedSet: TwinnedSet,
}
```

texastony: Oh this is an Enum Value?
Or something like that?
OK.

josecorella (approved):

PR #476, opened 2023-10-05
chore: update MPL submodule

josecorella:
*Issue #, if available:*

*Description of changes:*


DynamoDbEncryption/runtimes/net/DynamoDbEncryption.csproj
```diff
@@ -27,10 +27,10 @@
...
 
     <ItemGroup>
-        <ProjectReference Include="../../../submodules/MaterialProviders/AwsCryptographicMaterialProviders/runtimes/net/MPL.csproj" />
-        <ProjectReference Include="../../../submodules/MaterialProviders/AwsCryptographyPrimitives/runtimes/net/Crypto.csproj" />
-        <ProjectReference Include="../../../submodules/MaterialProviders/ComAmazonawsKms/runtimes/net/AWS-KMS.csproj" />
-        <ProjectReference Include="../../../submodules/MaterialProviders/ComAmazonawsDynamodb/runtimes/net/ComAmazonawsDynamodb.csproj" />
+        <PackageReference Include="AWS.Cryptography.MaterialProviders" Version="1.0.0" />
+        <PackageReference Include="AWS.Cryptography.Internal.AwsCryptographyPrimitives" Version="1.0.0" />
+        <PackageReference Include="AWS.Cryptography.Internal.ComAmazonawsDynamodb" Version="1.0.0" />
+        <PackageReference Include="AWS.Cryptography.Internal.ComAmazonawsKms" Version="1.0.0" />
```
texastony, line 33: Question: Should the STD be included?
The way Java is packaged, the STD is inside of the MPL.
But in .NET, it is not.
It is a separate library. 

I am not sure how .NET dependency resolution works.
But it's possible that the STD is not included at all.

codebuild/release/validate-release.yml
texastony, line 1: Note: We really need to make the Dafny URL a variable that is referenced.

ajewellamz: I agree.

DynamoDbEncryption/runtimes/net/DynamoDbEncryption.csproj
ajewellamz: I also don't know. I suppose we might find out as I try to make this release.

texastony: Put Another Way: 
The MPL depends on the other three libraries (KMS, DDB, Primitives).
If .NET had transitive dependency resolution, 
we would not need to declare these here,
since the MPL would bring them in.

But we do include these.

Thus, we either don't want to rely on a transitive dependency declaration,
or .NET is not satisfied with a transitive dependency providing symbols to the project.

Either way, we should add the STD.

ajewellamz: Added

texastony (approved):

PR #482, opened 2023-10-08
feat: dotnet examples

ajewellamz:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/DynamoDbEncryption/dafny/DynamoDbEncryptionTransforms/src/PutItemTransform.dfy
+++ b/DynamoDbEncryption/dafny/DynamoDbEncryptionTransforms/src/PutItemTransform.dfy
@@ -32,7 +32,7 @@ module PutItemTransform {
       //= type=implication
       //# The PutItem request MUST NOT refer to any legacy parameters,
       //# specifically Expected and ConditionalOperator MUST NOT be set.
-      && input.sdkInput.Expected.None? && input.sdkInput.ConditionalOperator.None?
+      && NoMap(input.sdkInput.Expected) && input.sdkInput.ConditionalOperator.None?
 
       // && var oldHistory := old(tableConfig.itemEncryptor.History.EncryptItem);
       // && var newHistory := tableConfig.itemEncryptor.History.EncryptItem;
@@ -60,12 +60,9 @@ module PutItemTransform {
     }
     var tableConfig := config.tableEncryptionConfigs[input.sdkInput.TableName];
 
-    if input.sdkInput.Expected.Some? {
-      return MakeError("Legacy parameter 'Expected' not supported in PutItem with Encryption.");
-    }
-    if input.sdkInput.ConditionalOperator.Some? {
-      return MakeError("Legacy parameter 'ConditionalOperator' not supported in PutItem with Encryption.");
-    }
+    :- Need(NoMap(input.sdkInput.Expected), E("Legacy parameter 'Expected' not supported in PutItem with Encryption."));
+    :- Need(input.sdkInput.ConditionalOperator.None?, E("Legacy parameter 'ConditionalOperator' not supported in PutItem with Encryption."));
+
     var _ :- IsWriteable(tableConfig, input.sdkInput.Item);
     var _ :- TestConditionExpression(tableConfig,
       input.sdkInput.ConditionExpression,
--- a/DynamoDbEncryption/runtimes/net/DynamoDbEncryption.csproj
+++ b/DynamoDbEncryption/runtimes/net/DynamoDbEncryption.csproj
@@ -23,6 +23,7 @@
 
         <Compile Include="Extern/**/*.cs" />
         <Compile Include="Generated/**/*.cs" />
+        <Compile Include="src/**/*.cs" />
         <Compile Include="ImplementationFromDafny.cs" />
     </ItemGroup>
 
--- a/Examples/runtimes/net/src/BasicPutGetExample.cs
+++ b/Examples/runtimes/net/src/BasicPutGetExample.cs
@@ -1,0 +1,158 @@
+using Amazon.DynamoDBv2;
+using Amazon.DynamoDBv2.Model;
+using System.Threading.Tasks;
+using System;
+using System.Collections.Generic;
+using AWS.Cryptography.DbEncryptionSDK;
+using AWS.Cryptography.DbEncryptionSDK.DynamoDb;
+using AWS.Cryptography.DbEncryptionSDK.StructuredEncryption;
+using AWS.Cryptography.MaterialProviders;
+using System.Diagnostics;
+using System.Net;
+using Dafny;
+
+/*
+  This example sets up DynamoDb Encryption for the AWS SDK client
+  and uses the low level PutItem and GetItem DDB APIs to demonstrate
+  putting a client-side encrypted item into DynamoDb
+  and then retrieving and decrypting that item from DynamoDb. 
+
+  Running this example requires access to the DDB Table whose name
+  is provided in CLI arguments.
+  This table must be configured with the following
+  primary key configuration:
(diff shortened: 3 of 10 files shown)


Examples/runtimes/net/src/TestUtils.cs
seebees, line 1: I assume that these are used somewhere.
Should we put this list in Dafny so that every runtime has access to the same strings?

Examples/runtimes/net/src/BasicPutGetExample.cs
```diff
@@ -0,0 +1,158 @@
...
+  primary key configuration:
+    - Partition key is named "partition_key" with type (S)
+    - Sort key is named "sort_key" with type (N)
+ */
+public class BasicPutGetExample {
+
+    public static async Task PutItemGetItem()
+    {
+        var kmsKeyId = TestUtils.TEST_KMS_KEY_ID;
+        var ddbTableName = TestUtils.TEST_DDB_TABLE_NAME;
```
lucasmcdonald3, line 32: Users should be able to pass in their own resource ARNs via a CLI args.

Two questions:
1. Could you have PutItemGetItem take in these arguments?
1. Could you add a main method to the bottom of this file that takes in command-line arguments and passes them to PutItemGetItem?

Examples/runtimes/net/src/TestUtils.cs
ajewellamz: It's the same list used in the Java example. 
I presume that once I've moved over all the examples, that all these will be used.
Putting them in a central place seems reasonable. I'm not sure Dafny is the right choice, but maybe I just don't see whatever you're envisioning.

Examples/runtimes/net/src/Examples.cs
```diff
@@ -0,0 +1,19 @@
+﻿
```
lucasmcdonald3, line 1: Users should be able to execute examples independently of one another.
I think you should remove this file in favor of per-example Main methods.
Also for parsing `args` and passing them into the example method.

Examples/runtimes/net/src/TestUtils.cs
```diff
@@ -0,0 +1,19 @@
+using System;
+public class TestUtils
+{
+    public static readonly string TEST_KEYSTORE_NAME = "KeyStoreDdbTable";
```
lucasmcdonald3, line 4: Can this go in a `runtimes/net/tests` or similar directory?
Then add a `Tests.csproj` file that executes the tests similar to https://github.com/aws/aws-encryption-sdk-dafny/blob/mainline/aws-encryption-sdk-net/Test/AWSEncryptionSDKTests.csproj

lucasmcdonald3 (commented):
I'm not looped in on our current plans for repos/languages, but:
Do we actually want to put this in the DDB-Java repo?

DynamoDbEncryption/dafny/DynamoDbEncryptionTransforms/src/PutItemTransform.dfy
```diff
@@ -32,7 +32,7 @@ module PutItemTransform {
       //= type=implication
       //# The PutItem request MUST NOT refer to any legacy parameters,
       //# specifically Expected and ConditionalOperator MUST NOT be set.
-      && input.sdkInput.Expected.None? && input.sdkInput.ConditionalOperator.None?
+      && NoMap(input.sdkInput.Expected) && input.sdkInput.ConditionalOperator.None?
```
texastony, line 35: Wow. If this is all that is needed to handle empty, we should change Java.
(But treat DDB's AttributeValue specially)

ajewellamz: For an optional map or list, the dotnet AWS SDK can't distinguish empty from None (since it hides the IsSet method).
This should be safe for java, since the smithy says "Optional non-empty".
I don't understand what you mean by "we should change java" -- as far as I can tell, Java is right and dotnet is wrong.

DynamoDbEncryption/runtimes/net/DynamoDbEncryption.csproj
```diff
@@ -1,46 +1,40 @@
...
-        <LangVersion>10</LangVersion>
-        <TargetFrameworks>net6.0;net48</TargetFrameworks>
-        <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
-    </PropertyGroup>
+  <PropertyGroup>
+    <RootNamespace>AWS Cryptography DynamoDb Encryption</RootNamespace>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>disable</Nullable>
+    <LangVersion>10</LangVersion>
+    <TargetFramework>net6.0</TargetFramework>
```
seebees, line 8: This would be a breaking change. I don't think we can do this.

DynamoDbEncryption/runtimes/net/DynamoDbEncryption.csproj
```diff
@@ -1,46 +1,40 @@
...
-        <PackageReference Include="System.ValueTuple" Version="4.5.0" />
-
-        <Compile Include="Extern/**/*.cs" />
-        <Compile Include="Generated/**/*.cs" />
-        <Compile Include="ImplementationFromDafny.cs" />
-    </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="AWSSDK.DynamoDBv2" Version="3.7.300.2"/>
+    <PackageReference Include="AWSSDK.Core" Version="3.7.300.2"/>
+    <ProjectReference Include="../../../submodules/MaterialProviders/AwsCryptographicMaterialProviders/runtimes/net/MPL.csproj" />
```
seebees, line 15: I don't think that this will work when the code is published.

TestVectors/runtimes/java/iotest.json
```diff
@@ -180,7 +111,7 @@
             ]
           },
           "Ten": {
-            "BS": []
+            "BS": [""]
```
seebees, line 114: Just wondering why change this?

seebees (changes requested):

TestVectors/runtimes/java/iotest.json
ajewellamz: Because empty sets are illegal. I don't know how is passed in Java.

DynamoDbEncryption/runtimes/net/DynamoDbEncryption.csproj
ajewellamz: At some point recently, it told me MPL didn't support net48. Maybe changing ProjectReference to PackageReference will fix that.

DynamoDbEncryption/runtimes/net/DynamoDbEncryption.csproj
ajewellamz: Good point. I'll switch it back.

seebees: @josecorella I think that the sub-module MPL here is only 6.0,
but I think that upstream is 6 and 4.8?

josecorella: yes, the mpl submodule commit on the dotnet-example is very out of date
https://github.com/aws/aws-cryptographic-material-providers-library-dafny/blob/13e0ac3c3c5eea83494706e4a96f40126d8f38a8/AwsCryptographicMaterialProviders/runtimes/net/MPL.csproj#L8

josecorella: @ajewellamz we should update the mpl submodule in this pr to point to the latest mpl changes

seebees (approved):
LGTM

PR #533, opened 2023-10-30
chore: update CFN template

seebees:
Include all repos so that CI
can use the test resources


lucasmcdonald3 (dismissed):
LGTM to deploy to CFN

texastony (approved):
We should assert this has been deployed before merging it.

texastony:
I verified that this has been deployed.

PR #609, opened 2023-11-27
feat: correct unicode less than

ajewellamz:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/DynamoDbEncryption/dafny/DynamoDbEncryption/test/FilterExpr.dfy
+++ b/DynamoDbEncryption/dafny/DynamoDbEncryption/test/FilterExpr.dfy
@@ -50,9 +50,7 @@ module TestDynamoDBFilterExpr {
     for i := 0 to |strs| {
       for j := 0 to |strs| {
         expect ((i < j) == UnicodeLess(strs[i], strs[j]));
-        if i != j {
-          expect ((i < j) == !UnicodeLess(strs[j], strs[i]));
-        }
+        expect ((i <= j) == !UnicodeLess(strs[j], strs[i]));
       }
     }
   }


DynamoDbEncryption/dafny/DynamoDbEncryption/test/FilterExpr.dfy
```diff
@@ -29,6 +29,25 @@ module TestDynamoDBFilterExpr {
...
+    var A := "A";
+    var B := "\ud000";
+    var C := "\ufe4c";
+    var D := "𐀂";
+    assert |A| == 1;
+    assert |B| == 1;
+    assert |C| == 1;
+    assert |D| == 2;
+
+    var strs := [A+B+C+D, C+D, D+C, D+C+B+A];
```
lavaleri, line 42: We should add a comment that this list is constructed to be ordered correctly. Otherwise it takes a moment to figure out what is actually under test here.

DynamoDbEncryption/dafny/DynamoDbEncryption/test/FilterExpr.dfy
```diff
@@ -29,6 +29,25 @@ module TestDynamoDBFilterExpr {
     expect does_contain(haystack, needle) == negate;
   }
 
+  method {:test} UnicodeLessTest() {
+    var A := "A";
+    var B := "\ud000";
```
lavaleri, line 34: I'm not sure how to feel about including what will amount to an invalid encoding in the test case... It will keep "undefined" behavior consistent, but may also send a signal that we will maintain this behavior.

DynamoDbEncryption/dafny/DynamoDbEncryption/test/FilterExpr.dfy
```diff
@@ -29,6 +29,25 @@ module TestDynamoDBFilterExpr {
...
+    var C := "\ufe4c";
+    var D := "𐀂";
```
lavaleri, line 36: I think tests should includes two more surrogate pairs:
- 1 surrogate pair that shares a high surrogate, but has a different low surrogate
- 1 surrogate pair that has a different high surrogate, may or may not have the same low surrogate.

DynamoDbEncryption/dafny/DynamoDbEncryption/test/FilterExpr.dfy
lavaleri: None of these test cases exercise the "only surrogates are involved" test case without already being equal.

DynamoDbEncryption/dafny/DynamoDbEncryption/test/FilterExpr.dfy
```diff
@@ -29,6 +29,25 @@ module TestDynamoDBFilterExpr {
...
+    assert |A| == 1;
+    assert |B| == 1;
+    assert |C| == 1;
+    assert |D| == 2;
+
+    var strs := [A+B+C+D, C+D, D+C, D+C+B+A];
+
+    for i := 0 to |strs| {
+      for j := 0 to |strs| {
+        expect ((i < j) == UnicodeLess(strs[i], strs[j]));
```
lavaleri, line 46: We should do the other direction as well. `expect (i < j) == !UnicodeLess(strs[j], strs[i])` Or prove it formally.

TestVectors/runtimes/java/data.json
```diff
@@ -1,4 +1,11 @@
 {
+  "Strings" : [
```
lavaleri, line 2: Can we add one more entry to ensure that the ordering is properly lexicographic? i.e. "퀀𐀂" goes right after "퀀"

DynamoDbEncryption/dafny/DynamoDbEncryption/test/FilterExpr.dfy
ajewellamz: I don't understand. d000 is a valid character.

DynamoDbEncryption/dafny/DynamoDbEncryption/src/FilterExpr.dfy
```diff
@@ -1262,19 +1262,36 @@ module DynamoDBFilterExpr {
...
+      if a[0] == b[0] then
+        UnicodeLess(a[1..], b[1..])
+      else if aIsSurrogate == bIsSurrogate then
+        a[0] < b[0]
+      else
+        bIsSurrogate
   }
 
-  predicate method LexicographicGreaterOrEqual<T(==)>(a: seq<T>, b: seq<T>, less: (T, T) -> bool)
+  predicate method UnicodeLessOrEqual(a : string, b : string)
```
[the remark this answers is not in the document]
ajewellamz: No, this is right.
`a <= b` is the same as `!(b < a)`

DynamoDbEncryption/dafny/DynamoDbEncryption/test/FilterExpr.dfy
lavaleri: Ah my bad, I misread as d800 so I thought it was a lone surrogate.

lavaleri (approved):

PR #797, opened 2024-03-07
chore: Adopt SmithyDafnyMakefile.mk, progress towards fixing nightly build

robin-aws:
*Description of changes:*

Replaces nearly all of the `SharedMakefile.mk` with the common `smithy-dafny/SmithyDafnyMakefile.mk` makefile, just retaining configuration variables specific to this repo (such as the path to the `smithy-dafny` submodule). Uses the new features in that makefile and `smithy-dafny` itself to make the projects forwards-compatible with the latest Dafny nightly prerelease, and hence will MOSTLY fix the nightly build once merged. "Mostly" because I still need to fix some externs to make them use the pattern that avoids the Java TypeDescriptor differences between Dafny versions, but that can be fixed in a follow-up PR.

Highlights of the changes:

* Apply the same workflow changes as https://github.com/aws/aws-cryptographic-material-providers-library/pull/195 to use `smithy-dafny` to regenerate code, either to check that the output matches what's checked in (in a new separate codegen workflow) or to be compatible with newer versions of Dafny in the nightly build (in existing workflows).
  * In this case we also have to locally update the MPL submodule to the latest, so that we can pick up the forwards-compatible changes to that repo, and regenerate code transitively. Generating code is unfortunately getting quite expensive, especially since the shared makefile logic isn't sophisticated enough to avoid regenerating the same code multiple times.
  * Because the code in this repo wasn't formatted already, but applying newer `smithy-dafny` code generation automatically formats, all the generated has trivial layout changes.
  * Also extracted a manual patch.
  * ~Applied a lot of [explicit client downcasting](https://github.com/aws/aws-database-encryption-sdk-dynamodb/pull/797/files#diff-692e2b06d124c9775e2fcd9cd9dbd10e0c8ea470e08174ed0b258b0301622581R182) to account for the change in `smithy-dafny` ~ (the change in smithy-dafny was undone so this isn't necessary any more)
* Converted `Beacon.CheckBytesToHex()` from a "test lemma" to a dynamic test, because on the latest Dafny the verification got even more expensive, and this didn't seem to introduce any real additional risk.


ajewellamz (approved):
LGTM

PR #819, opened 2024-03-14
feat: remove limits on list length and map size

ajewellamz:
*Description of changes:*

Remove the arbitrary limitations on list length and map size.
Keep the limit (32) on structure depth.

In earlier versions, it was thought that there might be a performance bottleneck in dealing with large maps and lists, but some code refinements and further testing showed that this was not the case, and that there was no need to limit customers in this way. Also, the 400K limit on an item puts an upper bound on structure complexity.


seebees (changes requested):
The code change looks good. But update the issue description to better explain why we think that this is safe etc. The description of the PR is what will go into the Changelog. So we can remove the issue link since one does not exist.

seebees (approved):

PR #972, opened 2024-04-29
chore: Update MPL to 1.3.0

ajewellamz:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.github/workflows/ci_test_vector_net.yml
+++ b/.github/workflows/ci_test_vector_net.yml
@@ -52,6 +52,8 @@ jobs:
         shell: bash
         working-directory: ./TestVectors
         run: |
+          # this is evil, but the MPL test vectors defines an extra Main
+          sed -i '' 's/method Main/method NotMain/' ../submodules/MaterialProviders/TestVectorsAwsCryptographicMaterialProviders/dafny/TestVectorsAwsCryptographicMaterialProviders/src/Index.dfy
           # This works because `node` is installed by default on GHA runners
           make transpile_net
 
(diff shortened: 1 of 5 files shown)


.github/workflows/ci_test_vector_net.yml
```diff
@@ -52,6 +52,9 @@ jobs:
         shell: bash
         working-directory: ./TestVectors
         run: |
+          # this is evil, but the MPL test vectors defines an extra Main
+          sed -i "s/method Main/method NotMain/" ../submodules/MaterialProviders/TestVectorsAwsCryptographicMaterialProviders/dafny/TestVectorsAwsCryptographicMaterialProviders/src/Index.dfy
+          sed -i "s/<OutputType>Exe/<OutputType>Library/" ../submodules/MaterialProviders/TestVectorsAwsCryptographicMaterialProviders/runtimes/net/TestVectors.csproj
```
seebees, line 57: can we just try actually renaming the method for now?
What happens to CI if we just do that?

texastony: This only happens in .NET.
I wonder if we changed the .csproj of the TestVectors to be a library, instead of an executable... 
Is that a thing in .NET?

ajewellamz: Yes, we're trying that in parallel. If it doesn't disable the test vectors, then we will update the MPL and remove those two sed commands.

ajewellamz: Change reverted, merged into MPL.

texastony (approved):

PR #1029, opened 2024-05-16
chore: Fix nightly build (aside from verification)

robin-aws:
*Description of changes:*

Makes the same changes around avoiding direct calls to instantiate parameterized datatypes in Java code as in https://github.com/aws/aws-cryptographic-material-providers-library/pull/195, plus some CI configuration fixes.

Dry runs of nightly builds:
* [Library Java tests](https://github.com/aws/aws-database-encryption-sdk-dynamodb/actions/runs/9104958211)
* [Java Examples](https://github.com/aws/aws-database-encryption-sdk-dynamodb/actions/runs/9106211948)
* [test dotnet](https://github.com/aws/aws-database-encryption-sdk-dynamodb/actions/runs/9100864042)

Will leave fixing the verification for a separate PR since this was already a fair bit of work :)


DynamoDbEncryption/runtimes/java/src/main/java/software/amazon/cryptography/dbencryptionsdk/dynamodb/itemencryptor/internaldafny/legacy/InternalLegacyOverride.java
```diff
@@ -13,6 +13,7 @@
  * IDEs tend to fight this so I'm sorry.
  */
 
+import Dafny.Aws.Cryptography.Primitives.Types.InternalResult;
```
texastony, line 16: Blocking/Issue: This is from the Primitives library, not the Standard Library or this Library (DB-ESDK).

That concerns me.
Can you explain why a Type defined in the Primitives Library is being used here?

I have a hard time believing that all the cases that you return a `InternalResult` always yield a class related to the Cryptographic Primitives Library.

texastony (changes requested):

DynamoDbEncryption/runtimes/java/src/main/java/software/amazon/cryptography/dbencryptionsdk/dynamodb/itemencryptor/internaldafny/legacy/InternalLegacyOverride.java
robin-aws: I introduced that `InternalResult` type when fixing up the primitives library, but it's just as generic as the Dafny `Result<T, E>` type: https://github.com/aws/aws-cryptographic-material-providers-library/blob/main/AwsCryptographyPrimitives/runtimes/java/src/main/java/Dafny/Aws/Cryptography/Primitives/Types/InternalResult.java#L3-L13

I didn't spot where I defined it when I imported it to solve the same problem here. It really should be in the Standard Library instead. But moving it there now in the MPL wouldn't fix the reference here until the next MPL launch so we can update the submodule.

I can just add a comment for now.

texastony: 1. I am happy I did not approve https://github.com/aws/aws-cryptographic-material-providers-library/pull/195. I get to continue to feel superior. 
2. an MPL release is going to go out very soon... go fix this now, and this PR can land correctly

seebees (approved):
LGTM

texastony (approved):

PR #1187, opened 2024-07-10
chore(GHA): add action for testing against MPL HEAD

josecorella:
*Issue #, if available:*

*Description of changes:*


texastony (approved):
Nice.

josecorella:
java test vectors will fail with this approach because the pinned version of the MPL 1.4.0 defines the test vector version as 1.0-SNAPSHOT while this pr updates the test vectors to depend on the same version for the mpl and the test vectors. The right thing to do here is to revert the change that changes the dependency definition of the test vectors so that we can merge this pr. The mpl-head-ci action failling daily is "ok" because it tells us that there needs to be project changes in order to merge with latest mpl head. This is *why* we wrote the action.

.github/workflows/ci_test_vector_java.yml
```diff
@@ -51,6 +60,20 @@ jobs:
...
+        working-directory: submodules/MaterialProviders
+        run: |
+          git checkout main
+          git pull
+          git submodule update --init --recursive
+          git rev-parse HEAD
+
+      - name: Update project.properties if using MPL HEAD
+        if: ${{ inputs.mpl-head == true }}
+        run: |
```
texastony, line 74:
```suggestion
      - name: Update MPL submodule if using MPL HEAD
        if: ${{ inputs.mpl-head == true }}
        working-directory: submodules/MaterialProviders
        run: |
          git checkout main
          git pull
          git submodule update --init --recursive
          git rev-parse HEAD

      - name: Update project.properties if using MPL Version
        if: ${{ inputs.mpl-version == true }}
        run: |
```

josecorella: we can't do this because mpl-version is of type string not boolean. This is why it comes after the Update MPL submodule if using MPL HEAD step

josecorella: it could be collapsed into one step; however, separating it into two makes it easier to debug

texastony:
> java test vectors will fail with this approach because the pinned version of the MPL 1.4.0 defines the test vector version as 1.0-SNAPSHOT while this pr updates the test vectors to depend on the same version for the mpl and the test vectors. The right thing to do here is to revert the change that changes the dependency definition of the test vectors so that we can merge this pr. The mpl-head-ci action failling daily is "ok" because it tells us that there needs to be project changes in order to merge with latest mpl head. This is _why_ we wrote the action.

When I was working on the KMS Exception issue,
one of the issues I struggled with was getting local changes picked up.

The build system relies on `mvn_local_deploy`.
This allows a locally build `TestVector` project to be utilized.
The issue I had was that I had a yee-old `TestVector` locally built and installed in my Maven Cache.

I believe that tying the Test Vector version to the MPL's version
ensures that all consumers update both at the same time.

Which will be necessary,
as the TestVectors will always need to be updated to consume the latest MPL,
or it will not be achieving it's objective of testing the MPL.

Recognizing that the TestVectors will always need to Version with the MPL,
it seems appropriate that TestVectors should have the same version.

This is not necessary; the TestVectors could follow any versioning number,
so long as that number is always incremented when the MPL's version is incremented.

But it feels unnecessary to have two separate numbers without a justification.

josecorella:
> > java test vectors will fail with this approach because the pinned version of the MPL 1.4.0 defines the test vector version as 1.0-SNAPSHOT while this pr updates the test vectors to depend on the same version for the mpl and the test vectors. The right thing to do here is to revert the change that changes the dependency definition of the test vectors so that we can merge this pr. The mpl-head-ci action failling daily is "ok" because it tells us that there needs to be project changes in order to merge with latest mpl head. This is _why_ we wrote the action.
> 
> When I was working on the KMS Exception issue, one of the issues I struggled with was getting local changes picked up.
> 
> The build system relies on `mvn_local_deploy`. This allows a locally build `TestVector` project to be utilized. The issue I had was that I had a yee-old `TestVector` locally built and installed in my Maven Cache.
> 
> I believe that tying the Test Vector version to the MPL's version ensures that all consumers update both at the same time.
> 
> Which will be necessary, as the TestVectors will always need to be updated to consume the latest MPL, or it will not be achieving it's objective of testing the MPL.
> 
> Recognizing that the TestVectors will always need to Version with the MPL, it seems appropriate that TestVectors should have the same version.
> 
> This is not necessary; the TestVectors could follow any versioning number, so long as that number is always incremented when the MPL's version is incremented.
> 
> But it feels unnecessary to have two separate numbers without a justification.

I might have not been clear in my comment blurb. I am not saying we need to keep track of two separate versions. At this time the DBESDK has not been bumped from 1.4.0 to 1.5.1

I do not want to bump in this PR. What I am saying is that the action mpl-head-ci is built to test against the latest changes of MPL, when it fails it means that there was a change in the MPL that this repo needs to update in order to build with the latest mpl such that we can bump to the latest released MPL. I left comments in the build.gradle.kts file to ensure we update the syntax.

texastony (approved):

.github/workflows/ci_test_java.yml
```diff
@@ -13,6 +13,15 @@ on:
...
         type: boolean
+      mpl-version:
+        description: "MPL version to use"
+        required: false
+        type: string
+      mpl-head:
+        description: "Running on MPL HEAD"
+        required: false
+        default: false
+        type: boolean
```
seebees, line 24: Is there a way to make it so you can only pass one?

.github/workflows/ci_test_java.yml
```diff
@@ -42,6 +51,20 @@ jobs:
         with:
           dafny-version: ${{ inputs.dafny }}
 
+      - name: Update MPL submodule if using MPL HEAD
+        if: ${{ inputs.mpl-head == true }}
```
seebees, line 55: What if head and version are true? Our ci is getting complicated :)

TestVectors/runtimes/java/build.gradle.kts
```diff
@@ -28,6 +28,7 @@ var dafnyRuntimeJavaVersion = props.getProperty("dafnyRuntimeJavaVersion")
 var smithyDafnyJavaConversionVersion = props.getProperty("smithyDafnyJavaConversionVersion")
 
 group = "software.amazon.cryptography"
+// change to ${mplVersion} for next MPL update
```
seebees, line 31: Not for this PR but can we put this under our release process?

seebees (commented):
Looks good

PR #1240, opened 2024-07-31
fix: allow multi-tenant queries with allow_plaintext

ajewellamz:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/DynamoDbEncryption/dafny/DynamoDbEncryptionTransforms/src/QueryTransform.dfy
+++ b/DynamoDbEncryption/dafny/DynamoDbEncryptionTransforms/src/QueryTransform.dfy
@@ -90,10 +90,10 @@ module QueryTransform {
       //# with the resulting decrypted [DynamoDB Item](./decrypt-item.md#dynamodb-item-1).
       var decryptInput := EncTypes.DecryptItemInput(encryptedItem := encryptedItems[x]);
       var decryptRes := tableConfig.itemEncryptor.DecryptItem(decryptInput);
-
       var decrypted :- MapError(decryptRes);
-      if keyId.KeyId? {
-        :- Need(decrypted.parsedHeader.Some?, E("Decrypted query result has no parsed header."));
+
+      // No parsed header is ok, because it means ALLOW_PLAINTEXT_READ and a plain text item
+      if keyId.KeyId? && decrypted.parsedHeader.Some? {
         :- Need(|decrypted.parsedHeader.value.encryptedDataKeys| == 1, E("Query result has more than one Encrypted Data Key"));
         if decrypted.parsedHeader.value.encryptedDataKeys[0].keyProviderInfo == keyIdUtf8 {
           decryptedItems := decryptedItems + [decrypted.plaintextItem];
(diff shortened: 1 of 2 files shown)


DynamoDbEncryption/dafny/DynamoDbEncryptionTransforms/src/QueryTransform.dfy
```diff
@@ -90,10 +90,10 @@ module QueryTransform {
       //# with the resulting decrypted [DynamoDB Item](./decrypt-item.md#dynamodb-item-1).
       var decryptInput := EncTypes.DecryptItemInput(encryptedItem := encryptedItems[x]);
       var decryptRes := tableConfig.itemEncryptor.DecryptItem(decryptInput);
-
       var decrypted :- MapError(decryptRes);
-      if keyId.KeyId? {
-        :- Need(decrypted.parsedHeader.Some?, E("Decrypted query result has no parsed header."));
+
+      // No parsed header is ok, because it means ALLOW_PLAINTEXT_READ and a plain text item
+      if keyId.KeyId? && decrypted.parsedHeader.Some? {
```
seebees, line 96: What is not clear to me is how `decrypted.parsedHeader.None?` will fail if `ALLOW_PLAINTEXT_READ` is false.

ajewellamz: No, the opposite.
If ALLOW_PLAINTEXT_READ is true, and the item read is plain text, then parsedHeader will be None.
In that case, we DO want to return the item, whereas the existing code will return an error.

ajewellamz: Or if you're asking something else:
decrypted.parsedHeader.None? <==> (ALLOW_PLAINTEXT_READ && plain text item)

lucasmcdonald3 (approved):

DynamoDbEncryption/dafny/DynamoDbEncryptionTransforms/src/ScanTransform.dfy
```diff
@@ -88,10 +88,19 @@ module ScanTransform {
...
+
+      // If the decrypted result was plaintext, i.e. has no parsedHeader
+      // then this is expected IFF the table config allows plaintext read
+      assert decrypted.parsedHeader.None? ==>
+          && EncOps.IsPlaintextItem(encryptedItems[x])
+          && !tableConfig.plaintextOverride.FORBID_PLAINTEXT_WRITE_FORBID_PLAINTEXT_READ?
+          && (
+               || tableConfig.plaintextOverride.FORBID_PLAINTEXT_WRITE_ALLOW_PLAINTEXT_READ?
+               || tableConfig.plaintextOverride.FORCE_PLAINTEXT_WRITE_ALLOW_PLAINTEXT_READ?
+             );
```
josecorella, line 101: nit: since this assert is the same as the one in QueryTransform, it would be nice to split into a predicate and just call it.

josecorella (approved):
a nice to have but we can always add later

PR #1244, opened 2024-08-02
chore: include bad item keys in query errors

ajewellamz:
*Issue #, if available:*

*Description of changes:*


seebees (approved):
This looks good.
We need an example since `CollectionOfErrors` has a complicated native interface.

PR #1391, opened 2024-09-25
chore(deps): Bump to MPL 1.7, Smithy-Dafny head

lucasmcdonald3:
*Issue #, if available:*

*Description of changes:*


josecorella (commented):
the part im still unclear on is why there are patch files needed for a mpl bump?
we have never had to do it before

josecorella (approved):

PR #1399, opened 2024-09-27
chore(deps): pin to always pull in the latest minor version

josecorella:
*Issue #, if available:*

*Description of changes:*


texastony (commented):
I would sooner put this in the properties file...

We can make `mplJavaVersion`,
and then declare this there as `1.+`.

I love the properties file...

PR #1474, opened 2024-11-08
fix: ensure algorithmSuite is properly copied in getTableConfig for E…

ajewellamz:
…nhancedClient

*Issue #, if available:*

*Description of changes:*


texastony (dismissed):
Anyway we could add a test for this?
Just so we know we keep this healthy in the future.

texastony (approved):

PR #1476, opened 2024-11-20
feat(SharedCache): Shared Cache for Searchable Encryption

RitvikKapila:
*Issue #, if available:*

*Description of changes:*
This PR:
- Adds support for a shared cache for Searchable Encryption beacons.
- Updates the cache identifiers for Searchable Encryption to be consistent with the MPL [Hierarchical Keyring identifier](https://github.com/awslabs/aws-encryption-sdk-specification/blob/master/framework/aws-kms/aws-kms-hierarchical-keyring.md#appendix-a-cache-entry-identifier-formulas) and [Caching CMM identifier](https://github.com/awslabs/aws-encryption-sdk-specification/blob/master/framework/caching-cmm.md#appendix-a-cache-entry-identifier-formulas).
- Updates the smithy model for `MultiKeyStore` to add an optional variable `partitionId`.
- Updates the smithy model for `SingleKeyStore` to add two optional variables: `partitionId` and `cache`.


DynamoDbEncryption/dafny/DynamoDbEncryption/src/ConfigToInfo.dfy
```diff
@@ -137,6 +138,7 @@ module SearchConfigToInfo {
       else
         MPT.Default(Default := MPT.DefaultCache(entryCapacity := 1));
 
+    // TODO : Add check that customers only provide either cacheTTL or cache in case of SingleKeyStore
```
RitvikKapila, line 141: Also for `MultiKeyStore`

DynamoDbEncryption/dafny/DynamoDbEncryption/Model/DynamoDbEncryption.smithy
```diff
@@ -713,9 +713,12 @@ structure SingleKeyStore {
   @required
   @javadoc("The Beacon Key ID.")
   keyId : String,
-  @required
-  @javadoc("How long (in seconds) the beacon key material is cached locally before it is re-retrieved from DynamoDB and re-authed with AWS KMS.")
+  @javadoc("How long (in seconds) the beacon key material is cached locally before it is re-retrieved from DynamoDB and re-authed with AWS KMS. Provide only one of cacheTTL or cache.")
```
RitvikKapila, line 716: `cacheTTL` should still be a required parameter, just like in the [MPL for Hierarchical Keyrings](https://github.com/aws/aws-cryptographic-material-providers-library/blob/1169bc8ef34ebfe3937ea3be8acf7a187ef1517a/AwsCryptographicMaterialProviders/dafny/AwsCryptographicMaterialProviders/Model/AwsCryptographyMaterialProvidersTypes.dfy#L909)

DynamoDbEncryption/dafny/DynamoDbEncryption/src/ConfigToInfo.dfy
texastony: I think this TODO should be removed, right?  
Because Cache's TTL is ALWAYS superseded by   
the cryptographic actor's TTL.

( cryptographic actor :: Searchable Encryption, CMM, Keyring, anything calling a CMC)

RitvikKapila: Yes, I updated the ticket but forgot to update this.

DynamoDbEncryption/dafny/DynamoDbEncryption/Model/AwsCryptographyDbEncryptionSdkDynamoDbTypes.dfy
```diff
@@ -388,7 +389,9 @@ module {:extern "software.amazon.cryptography.dbencryptionsdk.dynamodb.internald
   }
   datatype SingleKeyStore = | SingleKeyStore (
     nameonly keyId: string ,
-    nameonly cacheTTL: int32
+    nameonly cacheTTL: Option<int32> := Option.None ,
+    nameonly cache: Option<AwsCryptographyMaterialProvidersTypes.CacheType> := Option.None ,
```
RitvikKapila, line 393: Pass this shared cache like in `MultiKeyStore`

DynamoDbEncryption/dafny/DynamoDbEncryption/src/SearchInfo.dfy
```diff
@@ -253,6 +269,7 @@ module SearchableEncryptionInfo {
         return Failure(AwsCryptographyMaterialProviders(AwsCryptographyMaterialProviders:=getCacheOutput.error));
       }
```
RitvikKapila, line 270: Add this in the MPL HKeyring getActiveBranchKey and getVersionBranchKey to mitigate risk of swallowing errors
```
// If error is not EntryDoesNotExist, return failure
      if (getCacheOutput.Failure? && !getCacheOutput.error.EntryDoesNotExist?) {
        return Failure(AwsCryptographyMaterialProviders(AwsCryptographyMaterialProviders:=getCacheOutput.error));
      }
```

RitvikKapila: Also fix spec duvet for cache entry with limits where it is added

DynamoDbEncryption/dafny/DynamoDbEncryption/src/SearchInfo.dfy
```diff
@@ -241,8 +257,21 @@ module SearchableEncryptionInfo {
...
+      var scopeId : seq<uint8> := SCOPE_ID_SEARCHABLE_ENCRYPTION;
+
+      // Create the Suffix
       var keyIdBytesR := UTF8.Encode(keyId);
       var keyIdBytes :- keyIdBytesR.MapFailure(e => E(e));
+      var suffix : seq<uint8> := keyIdBytes;
+
+      // Append Resource Id, Scope Id, Partition Id, and Suffix to create the cache identifier
+      var identifier := resourceId + NULL_BYTE + scopeId + NULL_BYTE + partitionIdBytes + NULL_BYTE + suffix;
+
```
ajewellamz, line 274: I think the following line needs to be
`var getCacheInput := MP.GetCacheEntryInput(identifier := identifier, bytesUsed := None);`

RitvikKapila: You're right. Thanks Andy.

DynamoDbEncryption/dafny/DynamoDbEncryption/src/SearchInfo.dfy
```diff
@@ -199,22 +239,28 @@ module SearchableEncryptionInfo {
                 && var oldHistory := old(cache.History.GetCacheEntry);
                 && var newHistory := cache.History.GetCacheEntry;
                 && |newHistory| == |oldHistory|+1
-                && Seq.Last(newHistory).output.Success?
                 && var cacheInput := Seq.Last(newHistory).input;
                 && var cacheOutput := Seq.Last(newHistory).output;
                 && UTF8.Encode(keyId).Success?
-                && cacheInput.identifier == UTF8.Encode(keyId).value
```
RitvikKapila, line 245: @Ryan This is where I removed the cache Identifier check from.

DynamoDbEncryption/dafny/DynamoDbEncryption/src/SearchInfo.dfy
RitvikKapila: This is not an action item for this PR

RitvikKapila: Resolved here https://github.com/aws/aws-cryptographic-material-providers-library/pull/846

DynamoDbEncryption/dafny/DynamoDbEncryption/Model/DynamoDbEncryption.smithy
```diff
@@ -735,7 +742,9 @@ structure MultiKeyStore {
   @javadoc("How long (in seconds) the beacon key material is cached locally before it is re-retrieved from DynamoDB and re-authed with AWS KMS.")
   cacheTTL: Integer,
   @javadoc("Which type of local cache to use.")
```
seebees, line 744: I would keep the docs the same.

```suggestion
  @documentation("Provide the Shared Cache for Searchable Encryption.")
```

DynamoDbEncryption/dafny/DynamoDbEncryption/src/ConfigToInfo.dfy
```diff
@@ -38,6 +38,8 @@ module SearchConfigToInfo {
   method Convert(outer : DynamoDbTableEncryptionConfig)
     returns (output : Result<Option<I.ValidSearchInfo>, Error>)
     requires ValidSearchConfig(outer.search)
+    requires outer.search.Some? ==> ValidSharedCache(outer.search.value.versions[0].keySource)
+    modifies if outer.search.Some? then outer.search.value.versions[0].keyStore.Modifies else {}
```
seebees, line 42: These are 0, is there any requirement that `|outer.search.value.versions| == 1`?

DynamoDbEncryption/dafny/DynamoDbEncryption/src/ConfigToInfo.dfy
```diff
@@ -101,7 +117,9 @@ module SearchConfigToInfo {
   )
     returns (output : Result<I.KeySource, Error>)
     modifies client.Modifies
+    modifies keyStore.Modifies
     requires client.ValidState()
+    requires ValidSharedCache(config)
```
seebees, line 122: Why do you not `ensures ValidSharedCache(config)`? I'm not sure that it matters but I find that ensuring valid stat is just good hygiene.

DynamoDbEncryption/dafny/DynamoDbEncryption/src/ConfigToInfo.dfy
```diff
@@ -140,23 +146,38 @@ module SearchConfigToInfo {
...
         else
           MPT.Default(Default := MPT.DefaultCache(entryCapacity := 1000))
+      else
+      if config.single.cache.Some? then
+        // Ideally, we only want to pass a cache here with entryCapacity = 1
+        // because the SingleKeyStore caches only one value.
+        // That is, we SHOULD add a check here for entryCapacity = 1.
+        // However, that requires us to write an if block for each CacheType.
+        // Also, it does NOT matter what the entryCapacity is, because the cache
+        // can only hold one element at a time.
```
seebees, line 168: I'm not sure I agree.

Consider a system where I use a single config for multiple tables.
Each table has its own configuration
Just because this particular use will only every search for a single HMAC key,
does not mean that any other table used will search for one or even this particular key.

I could use this same CMC, to save allocations, in a `multi` config and expect the partition id
to keep the keys disjoint.

DynamoDbEncryption/dafny/DynamoDbEncryption/src/ConfigToInfo.dfy
```diff
@@ -140,23 +146,38 @@ module SearchConfigToInfo {
...
     var cache;
     if cacheType.Shared? {
-      return Failure(DynamoDbEncryptionException(message:="Searchable Encryption does not support the Shared Cache type at this time."));
-      // cache := cacheType.Shared;
+      cache := cacheType.Shared;
+      reveal ValidSharedCache(config);
+
+      // This axiom is important because it is not easy to prove
+      // keyStore.Modifies !! cache.Modifies for a shared cache.
+      assume {:axiom} keyStore.Modifies !! cache.Modifies;
```
seebees, line 180: What needs this to be true? Can we isolate this axiom a little more?

DynamoDbEncryption/dafny/DynamoDbEncryption/src/ConfigToInfo.dfy
```diff
@@ -168,20 +189,53 @@ module SearchConfigToInfo {
...
+      partitionIdBytes :- UTF8.Encode(config.single.partitionId.value)
+      .MapFailure(
+        e => Error.DynamoDbEncryptionException(
+            message := "Could not UTF-8 Encode Partition ID from SingleKeyStore: " + e
+          )
+      );
+    }
+    else {
+      partitionIdBytes :- I.GenerateUuidBytes();
+    }
```
seebees, line 212: nit: in dafny an `if` expression can be used for assignment.
```dafny
 var partitionIdBytes :- (if config.multi? && config.multi.partitionId.Some? then
    UTF8.Encode(config.multi.partitionId.value)
  else if config.single? && config.single.partitionId.Some? then
    UTF8.Encode(config.single.partitionId.value)
  else
    I.GenerateUuidBytes()
  )
  .MapFailure(
        e => Error.DynamoDbEncryptionException(
            message := "Could not UTF-8 Encode Partition ID from SingleKeyStore: " + e
          )
      );
```

You will likely have to do some type work if you want to change that, but you get the idea.

DynamoDbEncryption/dafny/DynamoDbEncryption/src/ConfigToInfo.dfy
```diff
@@ -168,20 +189,53 @@ module SearchConfigToInfo {
...
+    var getKeyStoreInfoOutput? := keyStore.GetKeyStoreInfo();
+    var getKeyStoreInfoOutput :- getKeyStoreInfoOutput?.MapFailure(e => Error.AwsCryptographyKeyStore(e));
+    var logicalKeyStoreName : string := getKeyStoreInfoOutput.logicalKeyStoreName;
```
seebees, line 215: Dafny's type inference it really good. Given that you don't do it everywhere, I would leave it off generally.

DynamoDbEncryption/dafny/DynamoDbEncryption/src/ConfigToInfo.dfy
```diff
@@ -168,20 +189,53 @@ module SearchConfigToInfo {
...
-      output := Success(I.KeySource(client, keyStore, I.SingleLoc(config.single.keyId), cache, config.single.cacheTTL as uint32));
+      output := Success(I.KeySource(client, keyStore, I.SingleLoc(config.single.keyId), cache, config.single.cacheTTL as uint32, partitionIdBytes, logicalKeyStoreNameBytes));
     }
   }
 
   // convert configured BeaconVersion to internal BeaconVersion
   method ConvertVersion(outer : DynamoDbTableEncryptionConfig, config : BeaconVersion)
     returns (output : Result<I.ValidBeaconVersion, Error>)
     requires ValidBeaconVersion(config)
+    requires ValidSharedCache(config.keySource)
```
seebees, line 237: same note about ensures

DynamoDbEncryption/dafny/DynamoDbEncryption/src/ConfigToInfo.dfy
```diff
@@ -168,20 +189,53 @@ module SearchConfigToInfo {
...
+    modifies config.keyStore.Modifies
```
seebees, line 238: Only keystore? Not the cache?

DynamoDbEncryption/dafny/DynamoDbEncryption/src/SearchInfo.dfy
```diff
@@ -127,6 +130,20 @@ module SearchableEncryptionInfo {
...
+  {
+    var uuid? := UUID.GenerateUUID();
+
+    var uuid :- uuid?
+    .MapFailure(e => Error.DynamoDbEncryptionException(message := e));
+
+    var uuidBytes: seq<uint8> :- UUID.ToByteArray(uuid)
+    .MapFailure(e => Error.DynamoDbEncryptionException(message := e));
+
+    output := Success(uuidBytes);
```
seebees, line 144: You can also

```dafny
output := UUID.ToByteArray(uuid)
    .MapFailure(e => Error.DynamoDbEncryptionException(message := e));
```

DynamoDbEncryption/dafny/DynamoDbEncryption/src/SearchInfo.dfy
```diff
@@ -153,17 +175,18 @@ module SearchableEncryptionInfo {
...
+        var theMap :- getKeysCache(client, stdNames, keyLoc.keyId, cacheTTL as MP.PositiveLong, partitionIdBytes, logicalKeyStoreNameBytes, now as MP.PositiveLong);
         return Success(Keys(theMap));
       } else if keyLoc.LiteralLoc? {
         :- Need(keyId.DontUseKeyId?, E("KeyID should not be supplied with a LiteralKeyStore"));
         var theMap :- getKeysLiteral();
         return Success(Keys(theMap));
       } else {
         match keyId {
-          case DontUseKeyId => return Failure(E("KeyID must not be supplied with a MultiKeyStore"));
+          case DontUseKeyId => return Failure(E("KeyID must be supplied with a MultiKeyStore"));
```
seebees, line 187: I get confused, @ajewellamz can you verify that this was a typeo?

DynamoDbEncryption/dafny/DynamoDbEncryption/src/SearchInfo.dfy
```diff
@@ -153,17 +175,18 @@ module SearchableEncryptionInfo {
...
           case ShouldHaveKeyId => return Success(ShouldHaveKeys);
-          case KeyId(id) => var theMap :- getKeysCache(stdNames, id); return Success(Keys(theMap));
+          case KeyId(id) => var now := Time.GetCurrent(); var theMap :- getKeysCache(client, stdNames, id, cacheTTL as MP.PositiveLong, partitionIdBytes, logicalKeyStoreNameBytes, now as MP.PositiveLong); return Success(Keys(theMap));
```
seebees, line 189: Long line, lets break it up
```suggestion
          case KeyId(id) =>
            var now := Time.GetCurrent();
            var theMap :- getKeysCache(
              client, stdNames, id,
              cacheTTL as MP.PositiveLong,
              partitionIdBytes,
              logicalKeyStoreNameBytes,
              now as MP.PositiveLong
            );
            return Success(Keys(theMap));
```

DynamoDbEncryption/dafny/DynamoDbEncryption/src/SearchInfo.dfy
```diff
@@ -231,29 +277,71 @@ module SearchableEncryptionInfo {
...
+      var identifier := resourceId + NULL_BYTE + scopeId + NULL_BYTE + partitionIdBytes + NULL_BYTE + suffix;
+
+      // Take a SHA384 of the cache identifier
+      var hashAlgorithm := Prim.DigestAlgorithm.SHA_384;
+
+      var identifierDigestInput := Prim.DigestInput(
+        digestAlgorithm := hashAlgorithm, message := identifier
+      );
+      var maybeCacheDigest := Digest.Digest(identifierDigestInput);
+      var cacheDigest :- maybeCacheDigest.MapFailure(e => AwsCryptographyPrimitives(e));
```
seebees, line 315: This looks like you are calling `Digest` explicitly. This is your problem. You should call this through the `client`.
I.e. `client.Digest`.

This will allow you to say something like
```dafny 
&& client.History.Digest[?].input.message == resourceId + NULL_BYTE + scopeId + NULL_BYTE + partitionIdBytes + NULL_BYTE + suffix`
&& client.History.Digest[?].output.Success?
&& cacheInput.identifier == client.History.Digest[?].output.value
```

DynamoDbEncryption/dafny/DynamoDbEncryption/src/SearchInfo.dfy
seebees: Commented below on how to do this.

DynamoDbEncryption/dafny/DynamoDbEncryption/src/SearchInfo.dfy
```diff
@@ -231,29 +277,71 @@ module SearchableEncryptionInfo {
...
+
+      :- Need(
+        |cacheDigest| == Digest.Length(hashAlgorithm),
+        Error.DynamoDbEncryptionException(
+          message := "Digest generated a message not equal to the expected length.")
+      );
```
seebees, line 321: This should be proved... See `DigestEnsuresPublicly`

DynamoDbEncryption/dafny/DynamoDbEncryption/src/SearchInfo.dfy
```diff
@@ -269,22 +357,21 @@ module SearchableEncryptionInfo {
         var keyMap :- getAllKeys(stdNames, key.value);
         var beaconKeyMaterials := rawBeaconKeyMaterials.beaconKeyMaterials.(beaconKey := None, hmacKeys := Some(keyMap));
 
+        expect now < UInt.BoundedInts.INT64_MAX - cacheTTL;
```
seebees, line 360: This is a reasonable `expect`. But understand, if this is not true, the caller does not get a nice recoverable error. This halts things. I don't know how hard in each runtime...

DynamoDbEncryption/dafny/DynamoDbEncryptionTransforms/src/Index.dfy
```diff
@@ -150,10 +151,11 @@ module
       var tableName: string := tableNamesSeq[i];
 
       var inputConfig := config.tableEncryptionConfigs[tableName];
-      :- Need(inputConfig.logicalTableName !in allLogicalTableNames,  E("Duplicate logical table maped to multipule physical tables: " + inputConfig.logicalTableName));
+      :- Need(inputConfig.logicalTableName !in allLogicalTableNames,  E("Duplicate logical table mapped to multiple physical tables: " + inputConfig.logicalTableName));
```
seebees, line 154: @ajewellamz  is it space bar heating for me to be worried about this? All we give customers at this time is the error string :(

seebees (changes requested):
still need to read the spec, but I think I showed how to prove things about the identifier

DynamoDbEncryption/dafny/DynamoDbEncryption/Model/DynamoDbEncryption.smithy
RitvikKapila: "Provide the Shared Cache for Searchable Encryption." for `MultiKeyStore` is misleading because customers can provide (and have been providing) a Default Cache for instance. This is NOT true for the `SingleKeyStore` because it never took a cache input. I am also inclined towards rejecting a non `Shared` cache for `SingleKeyStore`, just to avoid confusion.

DynamoDbEncryption/dafny/DynamoDbEncryption/src/ConfigToInfo.dfy
RitvikKapila: Yes, that is specified in a `:- Need` statement [here](https://github.com/aws/aws-database-encryption-sdk-dynamodb/blob/21f7fe168cedc0e6553926c7e3808081f172f7b0/DynamoDbEncryption/dafny/DynamoDbEncryption/src/ConfigToInfo.dfy#L58).

DynamoDbEncryption/dafny/DynamoDbEncryption/src/ConfigToInfo.dfy
RitvikKapila: I propose rejecting all cache inputs which are not shared.

DynamoDbEncryption/dafny/DynamoDbEncryption/src/ConfigToInfo.dfy
RitvikKapila: This is a great idea. But, I want to give different messages for `SingleKeyStore` and `MultiKeyStore`. To my knowledge, that won't work with this type of assignment.

DynamoDbEncryption/dafny/DynamoDbEncryption/src/ConfigToInfo.dfy
RitvikKapila: The cacheType inside config is converted into an actual cache inside `MakeKeySource`. It won't help to ensure ValidState of something we no longer need.

DynamoDbEncryption/dafny/DynamoDbEncryption/src/ConfigToInfo.dfy
RitvikKapila: That is a good point. I think I'd want this to be as is just so that its easier for someone who looks at this code to understand where the logicalKeyStoreName (or for that matter the partitionId) is a string and where it is converted into bytes (seq<uint8>).

DynamoDbEncryption/dafny/DynamoDbEncryption/src/ConfigToInfo.dfy
RitvikKapila: Great point. I don't think it is relevant here though. We can chat about this.

DynamoDbEncryption/dafny/DynamoDbEncryption/src/ConfigToInfo.dfy
RitvikKapila: I am calling the `GetKeyStoreInfo` for the underlying keyStore. The cache is either fresh, or it doesn't change (because it is `Shared`).

DynamoDbEncryption/dafny/DynamoDbEncryption/src/SearchInfo.dfy
RitvikKapila: Great point. Done! Thanks.

DynamoDbEncryption/dafny/DynamoDbEncryption/src/SearchInfo.dfy
RitvikKapila: I have pinged Andy to look at this. But you do need a KeyId to call `getKeyMap` for a `MultiKeyStore`. Think about it this way, before this change, the code for SingleKeyStore means:
I need the KeyId to be `DontUseKeyId` (which is expected because a SingleKeyStore only handles one KeyId, which it also holds. So you don't need to provide a KeyId.

On the other hand, for a MultiKeyStore, the code is:
If the KeyId is DontUseKeyId, return Failure saying "KeyID must not be supplied with a MultiKeyStore". But the irony is that we did NOT provide a KeyId.

DynamoDbEncryption/dafny/DynamoDbEncryption/src/SearchInfo.dfy
RitvikKapila: Great idea, I added the variables one at a time, and didn't realize when it got out of hand lol. Thanks for pointing this out.

DynamoDbEncryption/dafny/DynamoDbEncryption/Model/DynamoDbEncryption.smithy
RitvikKapila: ToDO(Ritvik): Write more words on doc

DynamoDbEncryption/dafny/DynamoDbEncryption/src/ConfigToInfo.dfy
RitvikKapila: TODO(Ritvik): Say that the cache is not Shared here.

DynamoDbEncryption/dafny/DynamoDbEncryption/src/SearchInfo.dfy
RitvikKapila: ToDo(Ritvik): Update in MPL.

DynamoDbEncryption/dafny/DynamoDbEncryption/Model/DynamoDbEncryption.smithy
RitvikKapila: Done

DynamoDbEncryption/dafny/DynamoDbEncryption/src/ConfigToInfo.dfy
RitvikKapila: Done. Thanks

DynamoDbEncryption/dafny/DynamoDbEncryption/src/ConfigToInfo.dfy
RitvikKapila: Done. Thanks

DynamoDbEncryption/dafny/DynamoDbEncryption/src/ConfigToInfo.dfy
RitvikKapila: Done. Thanks

DynamoDbEncryption/dafny/DynamoDbEncryption/src/ConfigToInfo.dfy
RitvikKapila: Done. Thanks

DynamoDbEncryption/dafny/DynamoDbEncryption/src/SearchInfo.dfy
RitvikKapila: I agree. I don't think there's any action item here.

DynamoDbEncryption/dafny/DynamoDbEncryptionTransforms/src/Index.dfy
ajewellamz: 1) We must be able to fix bugs and typos or all is lost.
2) It seems unlikely that anyone would write logic based the particular text of a "configuration error" message.

DynamoDbEncryption/dafny/DynamoDbEncryption/src/SearchInfo.dfy
ajewellamz, line 193:
```suggestion
          case DontUseKeyId => return Failure(E("Fixed KeyID must not be supplied with a MultiKeyStore"));
```

DynamoDbEncryption/dafny/DynamoDbEncryption/src/SearchInfo.dfy
RitvikKapila: Updated here: https://github.com/aws/aws-database-encryption-sdk-dynamodb/pull/1578

DynamoDbEncryption/dafny/DynamoDbEncryption/src/SearchInfo.dfy
RitvikKapila: Resolved

specification/searchable-encryption/search-config.md
```diff
@@ -332,3 +363,160 @@ using the beacon key retrieved above as the initial key material with no salt.
...
+
+Therefore, there are two important parameters that users need to carefully set while providing the shared cache:
+
+### Partition ID
+
+Partition ID is an optional parameter provided to the [Beacon Key Source](#beacon-key-source) input,
+which distinguishes Cryptographic Material Providers (i.e: [Beacon Key Sources](#beacon-key-source)) writing to a cache.
+
+- (Default) A a random 16-byte UUID, which makes
+  it unique for every [Beacon Key Source](#beacon-key-source). In this case, two [Beacon Key Sources](#beacon-key-source) (or another Material Provider)
```
seebees, line 498: There are a few places where newlines would be nice in the spec.
```suggestion
  it unique for every [Beacon Key Source](#beacon-key-source).
  In this case, two [Beacon Key Sources](#beacon-key-source) (or another Material Provider)
```

specification/searchable-encryption/search-config.md
```diff
@@ -332,3 +363,160 @@ using the beacon key retrieved above as the initial key material with no salt.
...
+  CANNOT share the same cache entries in the cache.
```
seebees, line 499: https://datatracker.ietf.org/doc/html/rfc2119 do you mean MUST NOT?

seebees: I think you mean will not.

specification/searchable-encryption/search-config.md
```diff
@@ -332,3 +363,160 @@ using the beacon key retrieved above as the initial key material with no salt.
...
+- If the Partition ID is set by the user and is the same for two [Beacon Key Sources](#beacon-key-source) (or another Material Provider),
+  they CAN share the same cache entries in the cache.
+- If the Partition ID is set by the user and is different for two [Beacon Key Sources](#beacon-key-source) (or another Material Provider),
+  they CANNOT share the same cache entries in the cache.
+
+### Logical Key Store Name
+
+> Note: Users MUST NEVER have two different physical Key Stores with the same Logical Key Store Name.
```
seebees, line 507: This is not a requirement that we can enforce.
And also, it is... nuanced. They can be the same, provided they can treat the physical stores as the same and ensure uniqueness.
Like a DDB global table. 

```suggestion
> Note: Users should not have two different physical Key Stores with the same Logical Key Store Name.
```

specification/searchable-encryption/search-config.md
```diff
@@ -332,3 +363,160 @@ using the beacon key retrieved above as the initial key material with no salt.
...
+
+Logical Key Store Name is set by the user when configuring the Key Store for
+the [beacon versions](#beacon-version-initialization). This is a logical name for the key store.
```
seebees, line 510:
```suggestion
the [beacon versions](#beacon-version-initialization).
This is a logical name for the key store.
```

specification/searchable-encryption/search-config.md
RitvikKapila: That makes sense. Updated. Thanks.

specification/searchable-encryption/search-config.md
RitvikKapila: Yes I mean WILL NOT. Thanks for pointing out.

specification/searchable-encryption/search-config.md
```diff
@@ -332,3 +363,160 @@ using the beacon key retrieved above as the initial key material with no salt.
...
+- Same Branch Key ID
+
+then they WILL share the cache entries in the `Shared` Cache.
+
+Any keyring that has access to the `Shared` cache MAY be able to use materials
+that it MAY or MAY NOT have direct access to.
+
+Users MUST make sure that all of Partition ID, Logical Key Store Name of the Key Store
+and Branch Key ID are set to be the same for two [beacon versions](#beacon-version-initialization) if and only they want the keyrings to share
+cache entries.
```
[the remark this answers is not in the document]
RitvikKapila: We cannot ensure this. I'll remove the MUST to make it should (not caps).

specification/searchable-encryption/search-config.md
```diff
@@ -332,3 +363,160 @@ using the beacon key retrieved above as the initial key material with no salt.
...
+### Preliminaries
+
+Each of the cache entry identifier formulas includes serialized information related to the branch key,
+as defined in the [Key Provider Info](../../submodules/MaterialProviders/aws-encryption-sdk-specification/framework/structures.md#key-provider-information).
+
+We establish the following definitions for the Cache Entry Identifier formula:
+
+#### Resource Identifier
+
+A Hex value that indicates if an element is from a Caching_CMM, Hierarchical_Keyring, or some other future resource.
```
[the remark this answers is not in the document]
RitvikKapila: Good catch. Done. Thanks.

specification/searchable-encryption/search-config.md
```diff
@@ -208,24 +216,33 @@ and used to extract a beacon key id from a query.
...
+equal to 1. If the user provides a cache which is not `Shared`, they SHOULD set the [Entry Capacity](../../submodules/MaterialProviders/aws-encryption-sdk-specification/framework/cryptographic-materials-cache.md#entry-capacity)
+of the provided `CacheType` to 1, because the [Single Key Store](#single-key-store-initialization) only ever caches one entry. Even if the user provides an entryCapacity > 1, the [Single Key Store](#single-key-store-initialization) will only cache one entry.
+For a [Multi Key Store](#multi-key-store-initialization), either the user provides a cache, or we create a cache that MUST have [Entry Capacity](../../submodules/MaterialProviders/aws-encryption-sdk-specification/framework/cryptographic-materials-cache.md#entry-capacity)
+equal to 1000.
+
+The Key Store Cache MUST be shared across different [Beacon Key Sources](#beacon-key-source) if and only if a `Shared` cache is used.
+In all other cases, the Key Store Cache MUST be bound to the [Beacon Key Source](#beacon-key-source).
```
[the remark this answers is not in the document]
RitvikKapila: I'll remove the caps. Thanks for pointing this out.

specification/searchable-encryption/search-config.md
```diff
@@ -320,6 +337,20 @@ These materials MUST be put into the associated [Key Store Cache](#key-store-cac
...
+[Put Cache Entry](../../submodules/MaterialProviders/aws-encryption-sdk-specification/framework/local-cryptographic-materials-cache.md#put-cache-entry)
+MUST be the same.
+
+If using a `Shared` cache across multiple [Beacon Key Sources](#beacon-key-source),
+different [Beacon Key Sources](#beacon-key-source) having the same `branchKey` can have different TTLs.
+In such a case, the expiry time in the cache is set according to the [Beacon Key Source](#beacon-key-source) that populated the cache.
+There MUST be a check (cacheEntryWithinLimits) to make sure that for the cache entry found, who's TTL has NOT expired,
+`time.now() - cacheEntryCreationTime <= ttlSeconds` is true and
+valid for TTL of the [Beacon Key Source](#beacon-key-source) getting the cache entry.
+If this is NOT true, then we MUST treat the cache entry as expired.
```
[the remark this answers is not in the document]
RitvikKapila: Right now its right after the Seachable Encryption Cache Identifier. I'll move it further up.

specification/searchable-encryption/search-config.md
```diff
@@ -320,6 +337,20 @@ These materials MUST be put into the associated [Key Store Cache](#key-store-cac
 with an [Expiry Time](../../submodules/MaterialProviders/aws-encryption-sdk-specification/framework/cryptographic-materials-cache.md#expiry-time)
 equal to now + configured [cacheTTL](#cachettl).
 
+The Searchable Encryption cache identifier for [Key Store Cache](#key-store-cache)
+[Get Cache Entry](../../submodules/MaterialProviders/aws-encryption-sdk-specification/framework/local-cryptographic-materials-cache.md#get-cache-entry)
+and the [Key Store Cache](#key-store-cache)
+[Put Cache Entry](../../submodules/MaterialProviders/aws-encryption-sdk-specification/framework/local-cryptographic-materials-cache.md#put-cache-entry)
+MUST be the same.
```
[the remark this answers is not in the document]
RitvikKapila: Yes, this does read better. Thanks!

specification/searchable-encryption/search-config.md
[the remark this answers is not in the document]
RitvikKapila: Updated

specification/searchable-encryption/search-config.md
```diff
@@ -208,24 +216,33 @@ and used to extract a beacon key id from a query.
...
-MUST be 1
-For a [Multi Key Store](#multi-key-store-initialization) the [Entry Capacity](../../submodules/MaterialProviders/aws-encryption-sdk-specification/framework/cryptographic-materials-cache.md#entry-capacity)
-MUST be key store's max cache size.
+For a [Single Key Store](#single-key-store-initialization), either the user provides a cache, or we create a cache that MUST have [Entry Capacity](../../submodules/MaterialProviders/aws-encryption-sdk-specification/framework/cryptographic-materials-cache.md#entry-capacity)
+equal to 1. If the user provides a cache which is not `Shared`, they SHOULD set the [Entry Capacity](../../submodules/MaterialProviders/aws-encryption-sdk-specification/framework/cryptographic-materials-cache.md#entry-capacity)
+of the provided `CacheType` to 1, because the [Single Key Store](#single-key-store-initialization) only ever caches one entry. Even if the user provides an entryCapacity > 1, the [Single Key Store](#single-key-store-initialization) will only cache one entry.
```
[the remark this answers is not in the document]
RitvikKapila: I was trying to put a note here about the comment we discussed about. But I don't think there's anything to enforce here so to avoid confusion, I'm just going to remove it. Thanks for pointing this out.

specification/searchable-encryption/search-config.md
```diff
@@ -208,24 +216,33 @@ and used to extract a beacon key id from a query.
...
-
 ### Key Store Cache
 
 For a Beacon Key Source a [CMC](../../submodules/MaterialProviders/aws-encryption-sdk-specification/framework/cryptographic-materials-cache.md)
 MUST be created.
-For a [Single Key Store](#single-key-store-initialization) the [Entry Capacity](../../submodules/MaterialProviders/aws-encryption-sdk-specification/framework/cryptographic-materials-cache.md#entry-capacity)
-MUST be 1
-For a [Multi Key Store](#multi-key-store-initialization) the [Entry Capacity](../../submodules/MaterialProviders/aws-encryption-sdk-specification/framework/cryptographic-materials-cache.md#entry-capacity)
-MUST be key store's max cache size.
+For a [Single Key Store](#single-key-store-initialization), either the user provides a cache, or we create a cache that MUST have [Entry Capacity](../../submodules/MaterialProviders/aws-encryption-sdk-specification/framework/cryptographic-materials-cache.md#entry-capacity)
```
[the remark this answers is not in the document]
RitvikKapila: This is how it is created right now. There are no other enforcing "Need" or "ensures" clauses, so I'll remove the caps. Saying something like, "either the user provides a cache, or we create a cache that has Entry Capacity equal to 1"

seebees (approved):
LGTM

PR #1478, opened 2024-11-26
fix(Java): Improve Collection of Errors string

texastony:
*Issue #, if available:*

*Description of changes:*
Collection of Errors is a frustrating error because it requires customer action
to completely serialize it into logs.

We can make a simple fix in Java to improve the CX,
by always serializing the list into the Collection Of Error's
message.

We concatenate the given error message with 
a serialization of all the nested errors.

_Squash/merge commit message, if applicable:_
```
fix(Java): Improve Collection of Errors string
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,6 @@ test/**/Output/*
 specification_compliance_report.html
 /compliance
 /.smithy.lsp.log
+
+# logs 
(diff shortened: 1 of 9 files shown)


seebees (approved):
This looks good. Thanks

PR #1518, opened 2024-12-18
fix: support new input validation

ajewellamz:
*Issue #, if available:*

*Description of changes:*


ajewellamz:
> I don't know Rust well. Is `?` a nullable indication?

No.

`let x = foo()?;`

means

```
let maybeX = foo();
if maybeX.is_failure() then return maybeX
let x = maybeX.value
```

RitvikKapila (approved):
LGTM

PR #1583, opened 2025-01-21
chore(CI): mpl head step for rust

josecorella:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.github/workflows/library_rust_tests.yml
+++ b/.github/workflows/library_rust_tests.yml
@@ -13,6 +13,11 @@ on:
         required: false
         default: false
         type: boolean
+      mpl-head:
+        description: "Running on MPL HEAD"
+        required: false
+        default: false
+        type: boolean
 
 jobs:
   testRust:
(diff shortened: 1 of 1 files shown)


.github/workflows/library_rust_tests.yml
```diff
@@ -59,6 +64,15 @@ jobs:
...
           dafny-version: nightly-2024-12-03-4954170
 
+      - name: Update MPL submodule if using MPL HEAD
+        if: ${{ inputs.mpl-head == true }}
+        working-directory: submodules/MaterialProviders
+        run: |
+          git checkout main
+          git pull
+          git submodule update --init --recursive
+          git rev-parse HEAD
```
rishav-karanjit, line 74: Question: Why do we need this? Is this to log current commit hash in CI?

rishav-karanjit (commented):
One question otherwise looks good to me

.github/workflows/library_rust_tests.yml
josecorella: when you do the normal submodule checkout in the set up above that only checks out the commit that is being tracked by the repo.
by going into the submodule and checking out the latest explicitly we make sure we pull all the latest changes including submodule changes within the mpl. 
the git rev-parse HEAD is a stdout message for us to visually check it is the latest.

rishav-karanjit (approved):
LGTM

PR #1605, opened 2025-01-30
chore(GHA): Run Java CI testing for MPL Latest Release

RitvikKapila:
*Issue #, if available:*

*Description of changes:*
Adds a daily job to check that the HEAD of DB-ESDK is compatible with the latest release of the MPL (pulled from Maven).

An important part is that this should be independent of Daily CI / Push CI / Pull CI, because when the MPL is actually updated, we will see errors with testing the last release, and we don't want our current CI process to get affected because of this.

We also won't need to go through the pain of running the MCM step to run [dafny-interop.yml](https://github.com/aws/aws-database-encryption-sdk-dynamodb/actions/workflows/dafny-interop.yml) manually for the latest released version of the MPL. This new workflow does it automatically, and the oncall just needs to verify that everything is green.


.github/workflows/ci_test_latest_released_mpl_java.yml
```diff
@@ -0,0 +1,32 @@
+# This workflow is for testing that the latest released version
+# of the MPL is compatible with the current DB-ESDK Head
+name: Test Latest Released MPL Java with DB-ESDK HEAD
+
+on:
+  schedule:
+    - cron: "00 16 * * 1-5"
```
josecorella, line 7: could we also make it run on demand?

.github/workflows/ci_test_latest_released_mpl_java.yml
```diff
@@ -0,0 +1,32 @@
...
+    if: github.event_name != 'schedule' || github.repository_owner == 'aws'
+    uses: ./.github/workflows/mpl_head_version.yml
+    with:
+      mpl-head: ${{needs.getLatestReleaseMplVersion.outputs.version}}
+  test-latest-released-mpl-java:
+    needs: [getDafnyVersion, getLatestReleaseMplVersion, getMplDafnyVersion]
+    uses: ./.github/workflows/dafny-interop.yml
+    with:
+      mpl-dafny: ${{needs.getLatestReleaseMplVersion.outputs.version}}
+      mpl-commit: ${{needs.getMplDafnyVersion.outputs.dafnyVersion}}
```
josecorella, line 31: this doesn't make sense, why for the mpl commit are we supplying the dafny version?

RitvikKapila: Because the MPL version will be built with its own Dafny version, and we pick this up from the project properties file of the MPL.

.github/workflows/ci_test_latest_released_mpl_java.yml
RitvikKapila: Done.

.github/workflows/ci_test_latest_released_mpl_java.yml
RitvikKapila: Look at the first input to [dafny-interop.yml](https://github.com/aws/aws-database-encryption-sdk-dynamodb/blob/main/.github/workflows/dafny-interop.yml).

.github/workflows/ci_test_latest_released_mpl_java.yml
```diff
@@ -0,0 +1,122 @@
+# This workflow is for testing that the latest released version
+# of the MPL is compatible with the current DB-ESDK Head
+name: Test Latest Released MPL Java with DB-ESDK HEAD
+
+on:
+  schedule:
+    - cron: "00 16 * * 1-5"
+  workflow_dispatch: # allows triggering this manually through the Actions UI
+  # TODO: Remove this before merging to main. We don't want this workflow running in CI
+  # on PRs because we expect it to fail when the MPL is actually updated since the last version.
```
RitvikKapila, line 10: Add a flag which by default is on, and if off, doesn't run 105-113 (Test Vectors) but runs the examples. Needed so that if TVs fail, we're still able to test examples.

.github/workflows/ci_test_latest_released_mpl_java.yml
```diff
@@ -0,0 +1,126 @@
...
+          gradle -p runtimes/java clean
+          gradle -p runtimes/java test
+
+      # This makes sure that we are using the correct MPL version to test the DB-ESDK.
+      # If this contains a SNAPSHOT version, this will fail because'
+      # we are NOT building the MPL recursively but pulling from Maven.
+      - name: Update project.properties to use the correct MPL version (from project.properties in DB-ESDK)
+        working-directory: ./submodules/MaterialProviders/
+        run: |
+          sed "s/mplVersion=.*/mplVersion=${{needs.getMplDependencyJavaVersion.outputs.version}}/g" project.properties > project.properties2; mv project.properties2 project.properties
```
josecorella, line 95: should this happen before we build and deploy to maven local on line 72?

RitvikKapila: The maven local deploy above this is only for DB-ESDK.

josecorella: right which will resolve its own dependencies.
So, here it will have mpl dependency x, and below in the test vectors it will have dependency y. 

To avoid getting in weird states let's change the version first and then build everything.

josecorella: ah reading fail

josecorella (approved):

PR #1861, opened 2025-05-05
feat(Go): DBESDK Go with some examples and some CI (#1719)

rishav-karanjit:
*Issue #, if available:*

*Description of changes:*

* chore(Go): Add DynamoDBEncryption Module 

* chore(Go): Add interop test with CI  (#1721)

* chore(Go): Add KMS and raw AES keyring example with CI (#1726)


DynamoDbEncryption/runtimes/go/ImplementationFromDafny-go/dbesdkmiddleware/middleware.go
```diff
@@ -0,0 +1,350 @@
...
+				return err
+			}
+			// Add response interceptor at the end of Finalize step
+			return stack.Finalize.Add(m.createResponseInterceptor(), middleware.After)
+		})
+	}
+}
+
+func (m DBEsdkMiddleware) createRequestInterceptor() middleware.InitializeMiddleware {
+	return middleware.InitializeMiddlewareFunc("RequestInterceptor", func(
```
ShubhamChaturvedi7, line 47: The ID `RequestInterceptor` is too generic, maybe add something like `aws-ddbec-request-interceptor` ?

DynamoDbEncryption/runtimes/go/ImplementationFromDafny-go/dbesdkmiddleware/middleware.go
```diff
@@ -0,0 +1,350 @@
...
+		ctx, err = m.handleRequestInterception(ctx, in.Parameters)
+		if err != nil {
+			return middleware.InitializeOutput{}, middleware.Metadata{}, err
+		}
+		return next.HandleInitialize(ctx, in)
+	})
+}
+
+// handleRequestInterception handles the interception logic before the DynamoDB operation
+func (m DBEsdkMiddleware) handleRequestInterception(ctx context.Context, request interface{}) (context.Context, error) {
```
ShubhamChaturvedi7, line 61: I assume operation that is not handled here is passed as it-is (given that this is global interceptor)

DynamoDbEncryption/runtimes/go/ImplementationFromDafny-go/dbesdkmiddleware/middleware.go
```diff
@@ -0,0 +1,350 @@
...
+	switch v := request.(type) {
+	case *dynamodb.PutItemInput:
+		ctx = middleware.WithStackValue(ctx, ContextKeyOriginalInput, *v)
+		// Note: this context is not propagated downstream into dafny layer so it's left as context.TODO() https://issues.amazon.com/CrypTool-5403
```
ShubhamChaturvedi7, line 65: Why not put the actual context here? So that if Dafny or some other tooling starts accepting it, we don't have to do anything here.

DynamoDbEncryption/runtimes/go/ImplementationFromDafny-go/dbesdkmiddleware/middleware.go
```diff
@@ -0,0 +1,350 @@
...
+		transformedRequest, err := m.client.PutItemInputTransform(context.TODO(), awscryptographydbencryptionsdkdynamodbtransformssmithygeneratedtypes.PutItemInputTransformInput{
+			SdkInput: *v,
+		})
+		if err != nil {
+			return nil, err
+		}
+		*v = transformedRequest.TransformedInput
```
ShubhamChaturvedi7, line 72: Q: if `v` is pointer type, I assume we are assigning value of `TransformedInput` to v and not address because we don't want to mutate the `TransformedInput` ?

DynamoDbEncryption/runtimes/go/ImplementationFromDafny-go/dbesdkmiddleware/middleware.go
```diff
@@ -0,0 +1,350 @@
...
+
+// createResponseInterceptor creates and returns the middleware interceptor for responses
+func (m DBEsdkMiddleware) createResponseInterceptor() middleware.FinalizeMiddleware {
+	return middleware.FinalizeMiddlewareFunc("ResponseInterceptor", func(
+		ctx context.Context, in middleware.FinalizeInput, next middleware.FinalizeHandler,
+	) (
+		out middleware.FinalizeOutput, metadata middleware.Metadata, err error,
+	) {
+		// First let the request complete
+		result, metadata, err := next.HandleFinalize(ctx, in)
```
ShubhamChaturvedi7, line 205: Doesn't `middleware.After` mean the interceptor will be executed after the request is done?

ShubhamChaturvedi7 (commented):
Mostly questions, otherwise looks good.

DynamoDbEncryption/runtimes/go/ImplementationFromDafny-go/dbesdkmiddleware/middleware.go
```diff
@@ -0,0 +1,350 @@
...
+		client: client,
+	}, nil
+}
+
+func (m DBEsdkMiddleware) CreateMiddleware() func(options *dynamodb.Options) {
+	return func(options *dynamodb.Options) {
+		options.APIOptions = append(options.APIOptions, func(stack *middleware.Stack) error {
+			// Add request interceptor at the beginning of Initialize step
+			requestIntercetor := m.createRequestInterceptor()
+			if err := stack.Initialize.Add(requestIntercetor, middleware.Before); err != nil {
```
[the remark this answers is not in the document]
rishav-karanjit: A request is created only after Serialize step which comes after Initialize step. Initialize step does the following:
- takes input parameters and processes them to ensure they're ready for the next steps in the pipeline.
- Sets default parameters. It fills in any default values that might be needed but weren't explicitly provided in the input.

So, putting a middleware stack before initialize step means that once the input come from application layer, the input will be encrypted then initialize step will get the ciphertext instead of the plaintext

DynamoDbEncryption/runtimes/go/ImplementationFromDafny-go/dbesdkmiddleware/middleware.go
rishav-karanjit: Yes, it is passed as it-is. We only handle 13 APIs for DB-ESDK because it makes sense to only have those and not have something like createTable. 

Say, DynamoDB adds a new API which makes sense to include here, we will have to manually update all the interceptors in all the runtimes in DB-ESDK. 

This is also what all other runtime does. Example (Java): https://github.com/aws/aws-database-encryption-sdk-dynamodb/blob/main/DynamoDbEncryption/runtimes/java/src/main/java/software/amazon/cryptography/dbencryptionsdk/dynamodb/DynamoDbEncryptionInterceptor.java#L326

DynamoDbEncryption/runtimes/go/ImplementationFromDafny-go/dbesdkmiddleware/middleware.go
rishav-karanjit: When I was writing this middleware I thought context.TODO() made more sense as we don't support context yet. But, as I think now, if we send the context now then we don't have do to anything if dafny starts accepting one.

DynamoDbEncryption/runtimes/go/ImplementationFromDafny-go/dbesdkmiddleware/middleware.go
rishav-karanjit: No. Its because we want to change content of `v` globally not just locally. `*v = transformedRequest.TransformedInput` means that it take a memory location that v points to and override the contents with the transformed input.

If we had done `v = &transformedRequest.TransformedInput`, this does not change the original v. We are just making a local variable v point to some different address and when the function returns the change to local variable is lost.

DynamoDbEncryption/runtimes/go/ImplementationFromDafny-go/dbesdkmiddleware/middleware.go
rishav-karanjit: When we get a response from transportation layer, it goes back to application layer in the reverse order of request (Deserialize -> Finalize -> Build -> Serialize -> Initialize). Adding after finalize means that after the final message (which will be a ciphertext) is prepared, we decrypt the message and send it to Build step.

DynamoDbEncryption/runtimes/go/ImplementationFromDafny-go/dbesdkmiddleware/middleware.go
rishav-karanjit: I will make a PR for this.

DynamoDbEncryption/runtimes/go/ImplementationFromDafny-go/dbesdkmiddleware/middleware.go
rishav-karanjit: I will make a PR for this.

DynamoDbEncryption/runtimes/go/ImplementationFromDafny-go/dbesdkmiddleware/middleware.go
rishav-karanjit, line 1: Filip comment: panic on default case

PR #1921, opened 2025-05-30
chore(dotnet): copy all the JSON for testvectors CI 

rishav-karanjit:
*Issue #, if available:*

*Description of changes:*
Current CI only copies `../java/decrypt_java_*.json ../java/decrypt_dotnet_*.json ../java/decrypt_rust_*.json ../java/large_records.json .` We do also need to copy files like `data.json`

### DIFF
--- a/.github/workflows/ci_test_vector_net.yml
+++ b/.github/workflows/ci_test_vector_net.yml
@@ -97,7 +97,6 @@ jobs:
       - name: Test TestVectors on .NET 6.0
         working-directory: ./${{matrix.library}}/runtimes/net
         run: |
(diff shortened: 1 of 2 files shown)


lucasmcdonald3 (approved):
Just wondering why this passes today before this change?
(This is still the right change)

rishav-karanjit:
> Just wondering why this passes today before this change? (This is still the right change)

rn, its just decrypting from the JSON that's already checked in but not creating its own JSON and doing round trip on it.
If |globalRecords| + |tableEncryptionConfigs| + |queries| == 0 then Dafny will just say Running no tests and let it pass. I am thinking talk with Andy if we can let it fail instead of printing Running no tests  and if there is any use case of not failing.
https://github.com/aws/aws-database-encryption-sdk-dynamodb/blob/main/TestVectors/dafny/DDBEncryption/src/TestVectors.dfy#L109

PR #1952, opened 2025-06-27
chore(cfn): add DescribeTable permission for legacy DDBEC support

imabhichow:
*Issue #, if available:*

*Description of changes:*
- This allows the DB-ESDK for Python to test the legacy extern behavior.
- See: https://docs.aws.amazon.com/database-encryption-sdk/latest/devguide/python-using.html#python-helpers
- To use the client helper classes in DDBEC, the caller must have permission to call the DynamoDB DescribeTable operation on the target table.

### DIFF
--- a/cfn/CI.yaml
+++ b/cfn/CI.yaml
@@ -256,6 +256,7 @@ Resources:
               - dynamodb:Scan
               - dynamodb:Query
               - dynamodb:UpdateItem
+              - dynamodb:DescribeTable
             Resource:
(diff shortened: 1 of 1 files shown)


texastony (changes requested):
This only grants Describe Table permission for the DynamoDB table used by the DB-ESDK's Interceptor. 

Your PR overview says you need it for the legacy table.  
Just add the action to the previous statement that covers all tables; that should cover both the legacy and contemporary tables.

texastony (approved):

PR #1966, opened 2025-08-01
chore(go): add plain text migration examples 

rishav-karanjit:
*Issue #, if available:*

*Description of changes:*
- Add examples to migrate from plain text table to DB-ESDK encrypted table.
- These tests is tested as a unit test but not by round tripping in main because we need to run various combination of stages of migration.
- Add utility functions required by examples in common utils file so that all examples can re-use same utils.
- Add README.md file 
- Some nits in existing Java example.

### DIFF
--- a/Examples/runtimes/go/migration/PlaintextToAWSDBE/awsdbe/step3.go
+++ b/Examples/runtimes/go/migration/PlaintextToAWSDBE/awsdbe/step3.go
@@ -1,11 +1,8 @@
 package awsdbe
 
 import (
-	// Standard imports
 	"context"
 	"fmt"
-
-	// AWS SDK imports
 
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/config"
@@ -14,6 +11,7 @@ import (
 
 	dbesdkdynamodbencryptiontypes "github.com/aws/aws-database-encryption-sdk-dynamodb/releases/go/dynamodb-esdk/awscryptographydbencryptionsdkdynamodbsmithygeneratedtypes"
 	"github.com/aws/aws-database-encryption-sdk-dynamodb/releases/go/dynamodb-esdk/dbesdkmiddleware"
+	plaintexttoawsdbe "github.com/aws/aws-database-encryption-sdk-dynamodb/releases/go/dynamodb-esdk/examples/migration/PlaintextToAWSDBE"
 	"github.com/aws/aws-database-encryption-sdk-dynamodb/releases/go/dynamodb-esdk/examples/utils"
 )
 
@@ -64,12 +62,14 @@ func MigrationStep3(kmsKeyID, ddbTableName, partitionKeyValue, sortKeyWriteValue
 	// 3. Put an item into your table.
 	//    This item will be encrypted.
 	encryptedAndSignedValue := "this will be encrypted and signed"
+	signOnlyValue := "this will never be encrypted, but it will be signed"
+	doNothingValue := "this will never be encrypted nor signed"
 	item := map[string]types.AttributeValue{
 		"partition_key": &types.AttributeValueMemberS{Value: partitionKeyValue},
 		"sort_key":      &types.AttributeValueMemberN{Value: sortKeyWriteValue},
 		"attribute1":    &types.AttributeValueMemberS{Value: encryptedAndSignedValue},
-		"attribute2":    &types.AttributeValueMemberS{Value: "this will never be encrypted, but it will be signed"},
-		"attribute3":    &types.AttributeValueMemberS{Value: "this will never be encrypted nor signed"},
+		"attribute2":    &types.AttributeValueMemberS{Value: signOnlyValue},
+		"attribute3":    &types.AttributeValueMemberS{Value: doNothingValue},
 	}
 
 	putInput := dynamodb.PutItemInput{
@@ -110,13 +110,10 @@ func MigrationStep3(kmsKeyID, ddbTableName, partitionKeyValue, sortKeyWriteValue
 	}
 
 	// Verify we got the expected item back
-	if partitionKeyValue != result.Item["partition_key"].(*types.AttributeValueMemberS).Value {
-		panic("Decrypted item does not match original item")
-	}
-	if encryptedAndSignedValue != result.Item["attribute1"].(*types.AttributeValueMemberS).Value {
-		panic("Decrypted item does not match original item")
+	err = plaintexttoawsdbe.VerifyReturnedItem(result, partitionKeyValue, sortKeyReadValue, encryptedAndSignedValue, signOnlyValue, doNothingValue)
+	if err != nil {
+		return err
 	}
 	fmt.Println("MigrationStep3 completed successfully")
-
 	return nil
 }


Examples/runtimes/go/migration/PlaintextToAWSDBE/awsdbe/step2_test.go
```diff
@@ -0,0 +1,34 @@
...
+	// Given: Step 0 has succeeded
+	plaintext.MigrationStep0(tableName, partitionKey, sortKeys[0])
+	// When: Execute Step 2 with sortReadValue=0, Then: Success (i.e. can read plaintext values)
+	MigrationStep2(kmsKeyID, tableName, partitionKey, sortKeys[0])
+
+	// Given: Step 1 has succeeded
+	MigrationStep1(kmsKeyID, tableName, partitionKey, sortKeys[1])
+	// When: Execute Step 2 with sortReadValue=1, Then: Success (i.e. can read encrypted values)
+	MigrationStep2(kmsKeyID, tableName, partitionKey, sortKeys[1])
+
```
imabhichow, line 29: Can we also add testcase to write with step 3 and read with step 2? also for `step 1_test.go`

see similar test in Python. https://github.com/aws/aws-database-encryption-sdk-dynamodb/blob/de4588f4dae01f659d6a2f14553b8b4c1efb686b/Examples/runtimes/python/Migration/plaintext_to_awsdbe/test/awsdbe/client/test_migration_step_2.py#L48-L56

rishav-karanjit: Thanks. It was a very good finding. 

I did it in this PR: [`4eb1d8a`](https://github.com/aws/aws-database-encryption-sdk-dynamodb/pull/1966/commits/4eb1d8a27e8a15dfbefe4948c055b05279de9a7b). I also figured that a change in local to handle error was not committed. So, this Commit include that too.

.github/workflows/ci_test_go.yml
```diff
@@ -140,8 +140,9 @@ jobs:
         run: |
           make test_go
 
-      - name: Test Examples
+      - name: Run and Test Examples
```
ShubhamChaturvedi7, line 143: Why run and test both? Since it's an example, I don't think we need to assert all cases via unit tests, but if it's already in then no worries.

Examples/runtimes/go/migration/PlaintextToAWSDBE/awsdbe/step1.go
```diff
@@ -0,0 +1,123 @@
...
+		"attribute2":    &types.AttributeValueMemberS{Value: "this will never be encrypted, but it will be signed"},
+		"attribute3":    &types.AttributeValueMemberS{Value: "this will never be encrypted nor signed"},
+	}
+
+	putInput := dynamodb.PutItemInput{
+		TableName: &ddbTableName,
+		Item:      item,
+	}
+
+	_, err = ddb.PutItem(context.TODO(), &putInput)
```
ShubhamChaturvedi7, line 82: Should we use a plain vanilla client to demonstrate that we actually aren't doing any encryption and can read unencrypted items using our client?

Examples/runtimes/go/migration/PlaintextToAWSDBE/awsdbe/common.go
```diff
@@ -0,0 +1,71 @@
...
+	// We will use the `CreateMrkMultiKeyring` method to create this keyring,
+	// as it will correctly handle both single region and Multi-Region KMS Keys.
+	matProv, err := mpl.NewClient(mpltypes.MaterialProvidersConfig{})
+	utils.HandleError(err)
+
+	keyringInput := mpltypes.CreateAwsKmsMrkMultiKeyringInput{
+		Generator: &kmsKeyID,
+	}
+	kmsKeyring, err := matProv.CreateAwsKmsMrkMultiKeyring(context.Background(), keyringInput)
+	utils.HandleError(err)
```
ShubhamChaturvedi7, line 25: Nice touch.

Examples/runtimes/go/migration/PlaintextToAWSDBE/awsdbe/step1.go
```diff
@@ -0,0 +1,123 @@
...
+
+	// We return this error because we run test against the error.
+	// When used in production code, you can decide how you want to handle errors.
+	if err != nil {
```
ShubhamChaturvedi7, line 86: Why not use utils here as well?

Examples/runtimes/go/migration/PlaintextToAWSDBE/awsdbe/step1.go
```diff
@@ -0,0 +1,123 @@
...
+		return err
+	}
+
+	// Verify we got the expected item back
+	if partitionKeyValue != result.Item["partition_key"].(*types.AttributeValueMemberS).Value {
+		panic("Decrypted item does not match original item")
+	}
+	if encryptedAndSignedValue != result.Item["attribute1"].(*types.AttributeValueMemberS).Value {
```
ShubhamChaturvedi7, line 118: What about other attributes? Maybe add comment if we don't want full assertion.

Examples/runtimes/go/migration/PlaintextToAWSDBE/awsdbe/step1_test.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+func TestMigrationStep1(t *testing.T) {
+	kmsKeyID := utils.KmsKeyID()
+	tableName := utils.DdbTableName()
+	partitionKey := uuid.New().String()
+	sortKeys := []string{"0", "1", "2", "3"}
+
+	// Successfully executes Step 1
+	err := MigrationStep1(kmsKeyID, tableName, partitionKey, sortKeys[1])
```
ShubhamChaturvedi7, line 18: Since we aren't mocking anything here, what's the difference between this and run?

Examples/runtimes/go/utils/exampleUtils.go
```diff
@@ -124,6 +133,33 @@ func HandleError(err error) {
 	}
 }
 
+func AssertServiceError(err error, service string, operation string, errorMessage string) {
+	if err == nil {
+		panic("Expected error but got no error")
```
ShubhamChaturvedi7, line 138: Don't we want to have this function as no-op in case of no errors? In that way, we can always call this instead of figuring out when to call this.

Examples/runtimes/go/utils/exampleUtils.go
```diff
@@ -137,3 +173,32 @@ func FileExists(filename string) bool {
...
+	partitionKeyName string,
+	partitionKeyValue string,
+	sortKeyName string,
+	sortKeyValue string,
+) {
+	cfg, err := config.LoadDefaultConfig(context.TODO())
+	HandleError(err)
+
+	// Create DynamoDB client
+	client := dynamodb.NewFromConfig(cfg)
```
ShubhamChaturvedi7, line 188: Or you could just use the already created client :)

Examples/runtimes/go/migration/PlaintextToAWSDBE/awsdbe/step1.go
rishav-karanjit: We do this in Test. In step 1 test, we write with vanilla DDB client and read with intercepted DDB client. In step 0 test, we write with intercepted DDB client and read with the vanilla one.

Examples/runtimes/go/migration/PlaintextToAWSDBE/awsdbe/step1.go
rishav-karanjit: Before writing migration example, I thought panic on error everywhere is good enough but with all the test combinations in migration test, it was not good enough. I think many in the team will still have a thought that we panic on every example. So, to make it visually distinct immediately visible in the code, I did not use utils.

Examples/runtimes/go/migration/PlaintextToAWSDBE/awsdbe/step1_test.go
rishav-karanjit: We run this only once with the test.

Examples/runtimes/go/utils/exampleUtils.go
rishav-karanjit: This functions is used where we don't have access of the ddb client. So, I am re-creating it. We also re-create ddb client in every example so that examples are easier to follow.

Examples/runtimes/go/migration/PlaintextToAWSDBE/awsdbe/step1.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	result, err := ddb.GetItem(context.TODO(), getInput)
+	// We return this error because we run test against the error.
+	// When used in production code, you can decide how you want to handle errors.
+	if err != nil {
+		return err
+	}
+
+	// Verify we got the expected item back
```
rishav-karanjit, line 114:
```suggestion
	// Verify we got the expected item back.
	// Since we are mostly concerned with Plaintext Override, 
	// our focus is to assert the encrypted and signed value.
```

Examples/runtimes/go/migration/PlaintextToAWSDBE/awsdbe/step1.go
rishav-karanjit: Will this change look good to you? https://github.com/aws/aws-database-encryption-sdk-dynamodb/pull/1966/files#r2255583436

I will apply this change to other places as well.

Examples/runtimes/go/migration/PlaintextToAWSDBE/awsdbe/step1.go
rishav-karanjit: We had decided yesterday to test every variable.

Examples/runtimes/go/migration/PlaintextToAWSDBE/awsdbe/step1.go
ShubhamChaturvedi7: We have a blocker on test but this PR is good to go: https://quip-amazon.com/dVbiAID0wjHq#DLY9DAFeZPq

ShubhamChaturvedi7 (approved):

PR #1985, opened 2025-09-06
java benchmark

rishav-karanjit:
*Issue #, if available:*

*Description of changes:*


texastony (changes requested):
I think this a good start, but that the AI generated some very weird things that we do not need.
There are many un-needed imports and un-needed README entries.

PR #1996, opened 2025-09-30
chore: remove aws sdk core dep

ShubhamChaturvedi7:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/DynamoDbEncryption/runtimes/net/DynamoDbEncryption.csproj
+++ b/DynamoDbEncryption/runtimes/net/DynamoDbEncryption.csproj
@@ -58,7 +58,6 @@
 
   <ItemGroup>
     <PackageReference Include="AWSSDK.DynamoDBv2" Version="3.7.402.2"/>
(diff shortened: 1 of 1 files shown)


DynamoDbEncryption/runtimes/net/DynamoDbEncryption.csproj
```diff
@@ -58,7 +58,6 @@
 
   <ItemGroup>
     <PackageReference Include="AWSSDK.DynamoDBv2" Version="3.7.402.2"/>
```
texastony, line 60: Also drop DDB. Let the MPL define it.

texastony (changes requested):

DynamoDbEncryption/runtimes/net/DynamoDbEncryption.csproj
ShubhamChaturvedi7: No we should keep DDB here, because DB-ESDK has a direct dependency on DDB.

PR #2048, opened 2026-01-30
feat(java)!: add DDBEC with SDK v2 and remove DDBEC with SDK V1 

rishav-karanjit:
*Issue #, if available:*

*Description of changes:*
BREAKING CHANGE: The AWS Database Encryption SDK for DynamoDB will NOT support AWS SDK for Java 1.x in the embedded 2.x version (which was known as DynamoDB Encryption Client(DDBEC)). The embedded DDBEC will now use AWS SDK for Java 2.x. These changes are limited to embedded DDBEC and rest of DB-ESDK has no code changes. 

If consumer of DB-ESDK is using APIs from namespace `com.amazonaws.services.dynamodbv2` through DB-ESDK then they have to migrate to use corresponding API from namespace `com.amazonaws.services.dynamodbv2.datamodeling.sdkv2`. If consumer of DB-ESDK are NOT using namespace `com.amazonaws.services.dynamodbv2` through DB-ESDK then there will not be any breaking change when moving to next major version.


josecorella (approved):
this is a rubber stamp and we had reviewed this large pr in smaller ones.

PR #2198, opened 2026-04-08
chore(ci): trigger daily CI in v3.x-java branch 

rishav-karanjit:
## How daily CI in v3.xJava Works?

1. **13:00 UTC weekday** — GitHub fires `daily_ci_java_3x.yml` on `main` (schedule only runs from default branch). This is just the dispatcher.

2. **The dispatcher job runs** — it calls the GitHub API:
   ```
   createWorkflowDispatch({
     workflow_id: 'daily_ci.yml',
     ref: 'v3.x-Java'
   })
   ```
   This tells GitHub: "trigger `daily_ci.yml` on branch `v3.x-Java`."

3. **GitHub receives the API call** — it looks at `daily_ci.yml` on the `v3.x-Java` branch and checks: "does this workflow accept `workflow_dispatch` as a trigger?" If yes, it starts a new workflow run. If no, the API call fails.

4. **GitHub starts a new run of `daily_ci.yml`** — this run is associated with `v3.x-Java`, not `main`. So when the jobs do `actions/checkout` (with no `ref`), they check out `v3.x-Java` code.

5. **Result**: Two separate workflow runs happen — the dispatcher (on `main`) and the real CI (on `v3.x-Java`). All reusable workflows run against `v3.x-Java` without any modifications.

### DIFF
--- a/.github/workflows/daily_ci_java_3x.yml
+++ b/.github/workflows/daily_ci_java_3x.yml
@@ -1,0 +1,43 @@
+name: Daily CI (main-dbesdk-java-v3.0)
+
+on:
+  schedule:
+    - cron: "00 13 * * 1-5"
+  workflow_dispatch:
+
+jobs:
+  trigger:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Fail after end of support
+        run: |
+          CUTOFF="2026-08-01"
+          NOW=$(date -u +%Y-%m-%d)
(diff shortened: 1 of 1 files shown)


.github/workflows/daily_ci_java_3x.yml
```diff
@@ -0,0 +1,48 @@
+name: Daily CI (main-dbesdk-java-v3.0)
```
rishav-karanjit:
```suggestion
name: Daily CI (v3.x-java)
```

rishav-karanjit:
Done in different way

PR #2203, opened 2026-04-09
chore(java): attempt to reduce flaky CI failures

sharmabikram:
*Issue #, if available:*

*Description of changes:*
TTL is set as 500 ms which is also the cache expiration time. However, there is a grace period in some cases in cache expiration time and hence we should wait for more than 500 ms in our tests. We do have other tests where it is waiting for 700 ms. This change aims to make sleep time consistent across.

Changes also sets fail fast to false to ensure all failures are caught at once.


rishav-karanjit (approved):
LGTM.

PR #2220, opened 2026-04-15
chore(java): Attempt to reduce flaky CI

sharmabikram:
*Issue #, if available:*

*Description of changes:*
1. .github/workflows/ci_test_vector_java.yml — Retry Docker image pull on macOS

The test vectors workflow uses Docker to run DynamoDB Local on macos-15-intel. Docker is installed via Colima, which 
has intermittent failures pulling container images (socket resets, layer corruption). Added an explicit docker pull 
step with 3 retries using nick-fields/retry@v3 before the rrainn/dynamodb-action runs. If the pull succeeds on any 
attempt, the action finds the image cached locally and skips the pull.

2. DynamoDbEncryptionInterceptorIntegrationTests.java — Fix TransactionConflict in parallel CI

TestTransactWriteAndGet used a hardcoded partition key "transact". When multiple CI matrix jobs (Java 8/11/17/19/21 × 
OS) run in parallel against the same DynamoDB table, concurrent transactWriteItems calls on the same item cause 
TransactionConflict. Fixed by appending the existing randomNum field (already used by other tests in this class for the
same reason) to the partition key.

3. cfn/CI.yaml — Switch CI tables to on-demand billing

### DIFF
--- a/.github/workflows/ci_test_vector_java.yml
+++ b/.github/workflows/ci_test_vector_java.yml
@@ -56,6 +56,14 @@ jobs:
         if: matrix.os == 'macos-15-intel' && matrix.library == 'TestVectors'
         uses: douglascamata/setup-docker-macos-action@v1.1.0
 
+      - name: Pre-pull DynamoDB Local image with retry
+        if: matrix.library == 'TestVectors'
+        uses: nick-fields/retry@v3
+        with:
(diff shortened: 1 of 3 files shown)


rishav-karanjit (approved):
LGTM

PR #2229, opened 2026-04-20
chore(java): retry failed workflow

sharmabikram:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.github/workflows/retry_daily_ci.yml
+++ b/.github/workflows/retry_daily_ci.yml
@@ -1,0 +1,40 @@
+# This workflow automatically re-runs failed jobs from the Daily CI.
+# It triggers once when the Daily CI completes, and if any jobs failed,
+# it re-runs only the failed jobs. This handles transient failures like
+# Maven Central outages, Docker pull errors, or credential issues.
+# It only retries once to avoid infinite loops.
+name: Retry Failed Daily CI
+
+on:
+  workflow_run:
+    workflows: ["Daily CI"]
+    types:
+      - completed
+
+jobs:
+  retry:
+    if: github.event.workflow_run.conclusion == 'failure'
+    runs-on: ubuntu-latest
+    permissions:
+      actions: write
+    steps:
+      - name: Re-run failed jobs
+        uses: actions/github-script@v7
+        with:
+          script: |
+            // Check if this run was already a retry to avoid infinite loops
+            const run = await github.rest.actions.getWorkflowRun({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              run_id: context.payload.workflow_run.id,
+            });
+            if (run.data.run_attempt > 1) {
+              console.log('Already a retry (attempt ' + run.data.run_attempt + '). Skipping.');
+              return;
+            }
+            console.log('Re-running failed jobs for run ' + context.payload.workflow_run.id);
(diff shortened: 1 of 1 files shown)


.github/workflows/retry_daily_ci.yml
rishav-karanjit, line 1: Blanket-retrying all failed jobs can silently mask real failures — a test that fails due to a genuine bug but passes on retry due to non-determinism would go unnoticed. I'd prefer we either scope retries to known-transient failure patterns (dependency resolution, docker pulls, credential issues) or evaluate each test to see if each of the test is non-deterministic and add retry for those test. As-is, this optimizes for green CI at the cost of CI trustworthiness.

sharmabikram: The next revision adds skip list functionality which can be used to not retry a given suite. Whenever we have fuzz tests in this library, we shall use that skip list. 
Except for dedicated fuzz tests, retry should be fine for general use.

.github/workflows/retry_daily_ci.yml
```diff
@@ -0,0 +1,74 @@
...
+            });
+            if (run.data.run_attempt > 1) {
+              console.log('Already a retry (attempt ' + run.data.run_attempt + '). Skipping.');
+              return;
+            }
+
+            // Jobs that should NOT be retried. These are non-deterministic tests
+            // (e.g., fuzz tests) where a retry could mask a real failure.
+            // Use job name prefixes/substrings to match.
+            const skipPatterns = [
```
rishav-karanjit, line 42: This assumes non deterministic job name will always include "fuzz". String matching on job name is too fragile.

.github/workflows/retry_daily_ci.yml
```diff
@@ -0,0 +1,74 @@
...
+            failedJobs.forEach(j => console.log(`  - ${j.name}`));
+
+            // Check if any failed job matches the skip list
+            const skipped = failedJobs.filter(job => {
+              return skipPatterns.some(pattern =>
+                job.name.toLowerCase().includes(pattern.toLowerCase())
+              );
+            });
+
+            if (skipped.length > 0) {
```
rishav-karanjit, line 63: If a unit test and a fuzz test fails, non of them will be retried. I prefer retrying unit test and skipping fuzz test.

.github/workflows/retry_daily_ci.yml
rishav-karanjit: I'd prefer an allowlist (opt in retries) over a denylist (opt out retries).

rishav-karanjit (approved):
I approved it. 

For the records: I'd prefer an allowlist (opt in retries) over a denylist (opt out retries). The yml does a sub string matching assuming 'fuzz' will be included in job name which is fragile. The current logic is if a unit test and a fuzz test fails, non of them will be retried. I prefer retrying unit test and skipping fuzz test.

.github/workflows/retry_daily_ci.yml
sharmabikram: If we are in situation where a any random test fails in first run and passes in second run masking the bug for the first run, we are doomed. This behavior MUST always be an exception or an explicit choice where we add fuzz tests. The skip list are generally made for exception cases and which is what is implemented here.

PR #2266, opened 2026-05-07
chore: use install smithy dafny deps from MPL 

rishav-karanjit:
*Issue #, if available:*

*Description of changes:*
We will reuse the action from MPL to avoid duplication across repositories to help us maintain the CI as we will have single source-of-truth.

Updated maintainers list in check-files.yml to add Bikram's username


sharmabikram (dismissed):

josecorella (dismissed):
this looks good, but we probably should have just updated the install_smithy_dafny action to use the mpl instead; because we had to touch every single workflow

sharmabikram (approved):

josecorella (approved):

PR #2351, opened 2026-06-30
fix(rust): use ERE-compatible regex in start_release.sh version check

josecorella:
The version check used the PCRE escape \d in an egrep pattern:

  REGEX_VERSION='^\d+\.\d+\.\d+$'
  MATCHES=$(echo "$1" | egrep $REGEX_VERSION | wc -l)

egrep uses POSIX Extended Regular Expressions, which do not support \d. The pattern therefore matched nothing, every input was treated as invalid, and the script always exited with:

  Version "X.Y.Z" must be N.N.N

This caused the 'Run start_release.sh and open a release PR' job to fail with exit code 1 even though VERSION=1.3.0 had already passed the workflow's own pre-check (which uses bash regex ^[0-9]+\.[0-9]+\.[0-9]+$).

Replace \d with [0-9] so the regex actually matches N.N.N, and quote the variable expansion in egrep for safety.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/DynamoDbEncryption/runtimes/rust/Cargo.toml
+++ b/DynamoDbEncryption/runtimes/rust/Cargo.toml
@@ -17,7 +17,7 @@ readme = "README.md"
 [dependencies]
 aws-config = "1.8.12"
 aws-lc-rs = {version = "1.17.0"}
-aws-lc-sys = { version = "0.41", optional = true }
+aws-lc-sys = { version = "0.42", optional = true }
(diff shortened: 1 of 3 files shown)


josecorella:
failing run: https://github.com/aws/aws-database-encryption-sdk-dynamodb/actions/runs/28477196474/job/84404666810

kessplas (dismissed):

kessplas (approved):

PR #2374, opened 2026-09-14
chore(test): DB-ESDK cross-language test-server language adapters

hanabanaka:
DB-ESDK cross-language test-server language adapters (java-v3, rust-v1, net-v4).

Companion to aws-crypto-tools-commons#16 (the test server, orchestrator, and test cases).

PR #2375, opened 2026-09-15
fix(rust): cap aws-smithy-types below 1.7 (aws-smithy-json 0.63 compat)

hanabanaka:
Cap `aws-smithy-types` below `1.7` in the Rust runtimes so `test_rust` keeps compiling.

`aws-smithy-types 1.7.0` changed `Document::Object` to carry a new `DocumentObject` struct instead of `HashMap<String, Document>`. The `aws-smithy-json 0.63.0` pulled in by the pinned `aws-sdk-*` / `aws-config` versions is written against the old shape and fails to compile against `1.7.0` (E0308 + E0004), which is currently red on `main` and every PR.

The direct pin was `aws-smithy-types = "1.3.6"` — a floor Cargo floats up to the newest `1.x`, so `1.7.0` silently got pulled in. Capping to `>=1.3.6, <1.7` keeps the runtimes on the `1.6.x` line (the resolution that built cleanly before `1.7.0` shipped). Applies to both `DynamoDbEncryption/runtimes/rust` and `TestVectors/runtimes/rust`.


hanabanaka:
Closing — no longer needed. The breakage was a transient window while `aws-smithy-types 1.7.0` was published but a compatible `aws-smithy-json` was not. `aws-smithy-json 0.64.0` (which supports `1.7.0`) has since been published, and because these runtimes carry no committed `Cargo.lock`, Cargo now re-resolves to the compatible `1.7.0 + 0.64.0` pair on its own.

Evidence: #2374, whose library runtime is byte-identical to `main` and does **not** include this cap, now passes `pr-ci-rust / testRust` (DynamoDbEncryption + TestVectors, ubuntu). The aws-smithy-json mismatch is a source-level compile error, so a green compile there means `main` compiles without this change.

This cap also pins `aws-smithy-types` *backward* onto the `1.6.x` line, the opposite direction from where the ecosystem healed. If the lockless resolution is a concern long-term, committing a `Cargo.lock` (or bumping the `aws-sdk-*` deps forward) is the better durable fix than a backward version cap.

